PR #4, opened 2015-06-05
Add quiet mode to only show full URL of found items.

averagesecurityguy:
Wanted to be able to use the output more easily when piping to other tools.

### DIFF
--- a/main.go
+++ b/main.go
@@ -54,6 +54,7 @@ type State struct {
 	Verbose        bool
 	UseSlash       bool
 	FollowRedirect bool
+	Quiet          bool
 	Mode           string
 	Printer        PrintResultFunc
 	Processor      ProcessorFunc
@@ -144,6 +145,7 @@ func ParseCmdLine() *State {
 	flag.StringVar(&extensions, "x", "", "File extension(s) to search for (dir mode only)")
 	flag.BoolVar(&s.Verbose, "v", false, "Verbose output (errors and IP addresses")
 	flag.BoolVar(&s.FollowRedirect, "r", false, "Follow redirects")
+	flag.BoolVar(&s.Quiet, "q", false, "Only print found items.")
 	flag.BoolVar(&s.UseSlash, "f", false, "Append a forward-slash to each directory request (dir mode only)")
 
 	flag.Parse()
@@ -348,7 +350,11 @@ func PrintDirResult(s *State, r *Result) {
 	if s.StatusCodes.Contains(r.Status) {
 		// Only print results out if we find something
 		// meaningful.
-		fmt.Printf("Found: /%s (%d)\n", r.Entity, r.Status)
+		if s.Quiet {
+			fmt.Printf("%s%s\n", s.Url, r.Entity)
+		} else {
+			fmt.Printf("Found: /%s (%d)\n", r.Entity, r.Status)
+		}
 	} else if s.Verbose {
 		// Print out other results if the user wants to
 		// see them.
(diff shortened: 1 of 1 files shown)


OJ:
Thanks @averagesecurityguy ! So this does seem to mix a few new features into one. Those features are:

* Quiet: This prevents printing of the banner or anything else that isn't a result.
* Full URL: This prints `http://.../...` instead of `Found /...`
* No status code: Disables the rendering of the status code

I like these options, but I think they should be split so that there are flags for each. I find the status code very useful. The use of `Found` vs `Result` is helpful when verbose mode is on.

Perhaps we should have it so that:

* `-q` : "Quiet" - removes all the noise (basically the banner at this point)
* `-e` : "Expanded" - Show URL rather than the shortened version which is just the URI
* `-n` : "No status codes" - prevents status codes from being rendered.
* We don't Render `Found` unless `-v` is enabled.

This means that the default behaviour without these flags would print the banner and show `/... (status)` for each element. Adding `-e` expands it to `http://.../... (status)`. Adding `-n` would mean `/...` only.

To get your behaviour, you'd add `-q -n -e` to your command line. How do you feel about that?

Thanks for the PR dude!

averagesecurityguy:
Sounds awesome. I can drop my original pull request and start separating
them out.

On Thursday, June 4, 2015, OJ Reeves <notifications@github.com> wrote:

> Thanks @averagesecurityguy <https://github.com/averagesecurityguy> ! So
> this does seem to mix a few new features into one. Those features are:
>
>    - Quiet: This prevents printing of the banner or anything else that
>    isn't a result.
>    - Full URL: This prints http://.../... instead of Found /...
>    - No status code: Disables the rendering of the status code
>
> I like these options, but I think they should be split so that there are
> flags for each. I find the status code very useful. The use of Found vs
> Result is helpful when verbose mode is on.
>
> Perhaps we should have it so that:
>
>    - -q : "Quiet" - removes all the noise (basically the banner at this
>    point)
>    - -e : "Expanded" - Show URL rather than the shortened version which
>    is just the URI
>    - -n : "No status codes" - prevents status codes from being rendered.
>    - We don't Render Found unless -v is enabled.
>
> This means that the default behaviour without these flags would print the
> banner and show /... (status) for each element. Adding -e expands it to
> http://.../... (status). Adding -n would mean /... only.
>
> To get your behaviour, you'd add -q -n -e to your command line. How do
> you feel about that?
>
> Thanks for the PR dude!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/OJ/gobuster/pull/4#issuecomment-109150376>.
>


-- 
--
Stephen Haywood
Owner, ASG Consulting
CISSP, OSCP
423.305.3700
asgconsulting.co

OJ:
I could also massage yours a bit myself. Just thought I'd run the changes by you first.

averagesecurityguy:
I'm great either way. Just let me know.

On Thursday, June 4, 2015, OJ Reeves <notifications@github.com> wrote:

> I could also massage yours a bit myself. Just thought I'd run the changes
> by you first.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/OJ/gobuster/pull/4#issuecomment-109154361>.
>


-- 
--
Stephen Haywood
Owner, ASG Consulting
CISSP, OSCP
423.305.3700
asgconsulting.co

OJ:
I'll merge this and adjust things in the process. Thanks :)

OJ:
Thanks for the PR @averagesecurityguy, I've made the adjustments, updated the documentation, etc. There should be an example of "grep mode" in the README. Thanks!

averagesecurityguy:
This looks great. Here's an example of my use case for grep mode.

https://gist.github.com/averagesecurityguy/b4da588d7553803128b0

It's still a work in progress but I wanted to run each of the fuzzdb predictable response paths against a single web server. Having the full URL makes it easier to create clickable links for each valid response.

OJ:
Yeah cool, gotcha. Well hopefully this will be good enough for you. It's something that should have been done for the original version, so thanks for the PR and the shove in the right direction!

averagesecurityguy:
It works perfectly. Next step is to hunt for web servers with Nmap or Nessus and feed a list of URLs to the bust.sh script.

cat url_list.txt | xargs -n 1 bust.sh

I find a lot of good in roads on pentests by looking at web servers. :)

OJ:
Great to hear! :)

PR #9, opened 2015-09-06
Sanity check prepending '.' on extension list. 

UID1K:
A small annoyance.

### DIFF
--- a/main.go
+++ b/main.go
@@ -331,7 +331,12 @@ func ProcessDirEntry(s *State, word string, resultChan chan<- Result) {
 
 	// Follow up with files using each ext.
 	for ext := range s.Extensions {
+		if strings.HasPrefix(s.Extensions[ext], ".") == false {
+			file := word + "." s.Extensions[ext]
+		}
+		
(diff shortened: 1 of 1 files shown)


OJ:
Ah! I see the story here! So you do `html,htm,php` instead of `.html,.htm,.php`.

Roger that. Thanks for the PR. I'll probably tweak this a little so that the prepending happens outside of the loop. Micro-optimisations for the win! But every bit I can take out of the loop, I will :)

OJ:
Hi there. In future can you please PR from feature branches? Thanks.

OJ:
Hi again,

Please see https://github.com/OJ/gobuster/compare/add-dots?expand=1 for an adjusted implementation. Instead of your approach it does the following:

* Checks for the `.` using a char compare instead of a string prefix check.
* Prepends the `.` during the startup rather than doing it every time a word in the wordlist is attempted.

Please take a look at let me know if this works for you. Thanks!

OJ:
I've added the fix to master.
Closing.

PR #13, opened 2016-01-27
Add "page length" (close #6)

### LINKED ISSUE #6: Add "page length" support
Add an option that will change the display to include the full length of the returned result (in bytes). This could be done via the `Content-Length` header or via parsing the page directly.

g0tmi1k:
As @OJ likes his memes....

... I had this going on in my head for the whole time.

### DIFF
--- a/main.go
+++ b/main.go
@@ -24,6 +24,8 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"io/ioutil"
+	"unicode/utf8"
 )
 
 // A single result which comes from an individual web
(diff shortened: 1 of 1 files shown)


OJ:
I'm going to tweak this just a little bit ;) But thanks for the submission!

I'll be adding `-l` as a parameter which allows you to include the length of the page in the output. If people don't want to know the length, then they don't have to suffer the overhead of it reading the page in to get it.

OJ:
I'm also going to make use of the http response's `ContentLength` property if it's set.

g0tmi1k:
I guessed you would be altering it - but at the very least - it has some functionality in to get the ball rolling ;).
Will go back and check out your commit, cos I had issues trying to read in `content-length` nicely.
Nice idea with `-l` =)

PR #14, opened 2016-01-27
Verbose DNS mode to show missing values

g0tmi1k:
...Prob not the best way to go about doing this (open to suggestions).

### DIFF
--- a/main.go
+++ b/main.go
@@ -358,6 +358,12 @@ func ProcessDnsEntry(s *State, word string, resultChan chan<- Result) {
 			result.Extra = strings.Join(ips, ", ")
 		}
 		resultChan <- result
+	} else if s.Verbose {
+		result := Result {
+			Entity: subdomain,
+			Status: 404,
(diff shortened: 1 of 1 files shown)


OJ:
It's cheeky, but I think it's fine :) I'll add a comment to the line so it's a bit clearer.

OJ:
Actually, I'm having second thoughts about this. I like verbose mode to show me the IPs, but I don't want the output to show me all of the missing DNS names at the same time.

Might have to change this up so that verbose does what it does with `dir` mode and add another flag to show IPs.

OJ:
I've merged this, but also doctored is as part of https://github.com/OJ/gobuster/commit/d54b3c44b3d54aa87bb80b404cb95d238f09341f -- not sure why it didn't auto close as "merged".

Cheers!

g0tmi1k:
Good calling!
Closing rather than merging helps to cover up my mistakes ;)

PR #17, opened 2016-05-17
Allows for empty lines in URL wordlists

g0tmi1k:
...because not every wordlist is 'clean'!


OJ:
Thanks @g0tmi1k! Nice catch. I'll be working this into the mainline shortly.

PR #28, opened 2016-09-25
Added functionality to save results to a file AND print CNAME records in DNS mode

knapsy:
Added ability to save results to a file with '-o' option. If no option provided, still defaults to STDOUT.

Also added ability to print CNAME records for identified subdomains in DNS mode ('-C' option). Could be useful for identifying "forgotten" subdomains that could potentially be used in "subdomain hijacking" attacks. Does not work with '-i' option (i.e. can either print IPs or CNAMEs, but not both).

### DIFF
--- a/main.go
+++ b/main.go
@@ -77,6 +77,7 @@ type State struct {
 	Quiet          bool
 	Setup          SetupFunc
 	ShowIPs        bool
+	ShowCNAME      bool
 	StatusCodes    IntSet
 	Threads        int
 	Url            string
@@ -85,6 +86,8 @@ type State struct {
 	Username       string
 	Verbose        bool
 	Wordlist       string
+	OutputFileName string
+	OutputFile     *os.File
 	IsWildcard     bool
 	WildcardForced bool
 	WildcardIps    StringSet
@@ -239,6 +242,7 @@ func ParseCmdLine() *State {
 	flag.StringVar(&s.Mode, "m", "dir", "Directory/File mode (dir) or DNS mode (dns)")
 	flag.StringVar(&s.Wordlist, "w", "", "Path to the wordlist")
 	flag.StringVar(&codes, "s", "200,204,301,302,307", "Positive status codes (dir mode only)")
+	flag.StringVar(&s.OutputFileName, "o", "", "Output file to write results to (defaults to stdout)")
 	flag.StringVar(&s.Url, "u", "", "The target URL or Domain")
 	flag.StringVar(&s.Cookies, "c", "", "Cookies to use for the requests (dir mode only)")
 	flag.StringVar(&s.Username, "U", "", "Username for Basic Auth (dir mode only)")
@@ -248,6 +252,7 @@ func ParseCmdLine() *State {
 	flag.StringVar(&proxy, "p", "", "Proxy to use for requests [http(s)://host:port] (dir mode only)")
 	flag.BoolVar(&s.Verbose, "v", false, "Verbose output (errors)")
 	flag.BoolVar(&s.ShowIPs, "i", false, "Show IP addresses (dns mode only)")
+	flag.BoolVar(&s.ShowCNAME, "C", false, "Show CNAME records (dns mode only, cannot be used with '-i' option)")
 	flag.BoolVar(&s.FollowRedirect, "r", false, "Follow redirects")
 	flag.BoolVar(&s.Quiet, "q", false, "Don't print the banner and other noise")
 	flag.BoolVar(&s.Expanded, "e", false, "Expanded mode, print full URLs")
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -570,12 +594,20 @@ func ProcessDirEntry(s *State, word string, resultChan chan<- Result) {
...
 	} else if s.ShowIPs {
-		fmt.Printf("Found: %s [%s]\n", r.Entity, r.Extra)
+		output += fmt.Sprintf("Found: %s [%s]\n", r.Entity, r.Extra)
+	} else if s.ShowCNAME {
+		output += fmt.Sprintf("Found: %s [%s]\n", r.Entity, r.Extra)
 	} else {
-		fmt.Printf("Found: %s\n", r.Entity)
+		output += fmt.Sprintf("Found: %s\n", r.Entity)
+	}
+	fmt.Printf(output)
```
OJ: Ooh format string attack vector! :D I think you meant `fmt.Println` here dude. All good I can fix this on merge.

OJ:
Thanks @knapsy ! Appreciate the effort mate. Do you have as sample run handy?

I'll get around to looking at landing this in a couple of days. Cheers bud!

OJ:
For future ref dude, can you please PR from a feature branch, rather than from `master`? Cheers man :) #pickybastard

main.go
knapsy: Ahh, should have known better - good pick up! :)

knapsy:
Sure thing, will do mate, "feature" branch from now it is 👍 

I've got few sample runs, will pull them out and attach here if you like?

OJ:
Yeah that'd be sweet. Thanks mate!

knapsy:
Here we go, 'output.txt' file is just STDOUT, I've attached the output file as well. Killed it after few seconds of running, but the output file is written on the fly anyway.

[output.txt](https://github.com/OJ/gobuster/files/491582/output.txt)
[yahoo-dns.txt](https://github.com/OJ/gobuster/files/491584/yahoo-dns.txt)

main.go
knapsy: Actually, I did some reading and it looks like `fmt.Println` will not work here as it doesn't return `string`, so we won't be able to save the value to a variable here. `fmt.Println` actually returns 2 values - `int` of number of bytes written and `err` - [https://golang.org/pkg/fmt/#Println](https://golang.org/pkg/fmt/#Println).

My quick Google-fu did not yield any simple solutions without using `fmt.Sprintf`, may need to go with this ugliness below instead?
`output += "Found: " + r.Entity + "\n"`

OJ:
I wasn't looking to replace Sprintf mate. Just the Printf call.

knapsy:
but then we can't nicely pass the `output` string to the `WriteToFile` function a bit further down?

main.go
OJ: OK so to be clear: I don't want to replace the `Sprintf`. If the newlines are already baked into the string, then instead of using `Println` we can use `fmt.Printf("%s", output)`, but we're not going to leave it as `fmt.Printf(output)` :)

knapsy: Ahhhhh ok, I see what you meant now! Yup, that makes perfect sense :) Glad we finally got there! :+1:

OJ:
Hey @knapsy where did we go to with this mate? Sorry I've been dark!

knapsy:
Hey mate, all good, completely forgot about it too.

Umm, stretching my memory a bit here, but I just had a look at the code again and I think that apart from that format string vulnerability (line 607), which should instead be `fmt.Printf("%s", output)` I think it should be all good, what do you reckon?

OJ:
Seems reasonable to me! If you're cool to make the change, I'll get this landed by the 28th for v1.3.

Cheers!

main.go
```diff
@@ -248,6 +252,7 @@ func ParseCmdLine() *State {
 	flag.StringVar(&proxy, "p", "", "Proxy to use for requests [http(s)://host:port] (dir mode only)")
 	flag.BoolVar(&s.Verbose, "v", false, "Verbose output (errors)")
 	flag.BoolVar(&s.ShowIPs, "i", false, "Show IP addresses (dns mode only)")
+	flag.BoolVar(&s.ShowCNAME, "C", false, "Show CNAME records (dns mode only, cannot be used with '-i' option)")
```
OJ: any chance we can rename `-C` to `-cn` please?

main.go
```diff
@@ -482,6 +498,9 @@ func Process(s *State) {
 	processorGroup.Wait()
 	close(resultChan)
 	printerGroup.Wait()
+	if s.OutputFileName != "" {
```
OJ: Might also be worth checking `s.OutputFile` for `nil` rather than relying on the name.

knapsy: yep, I agree! Would you like me to push down the changes (s.outputFileName and -cn flag)?

OJ: Yeah mate, please commit changes to the same branch, and just push them up. The PR will update.

knapsy:
All done... I think.

No idea why it came up as 'invalid-email-address' author, my git game is not strong :-|

OJ:
It's in the v1.3 branch, will be in master soon.

OJ:
It's coz you suck, but that's ok :)

OJ:
I am, of course, kidding 👍

knapsy:
ha-ha, <3 you too :P

PR #30, opened 2016-10-07
Implemented switch for URL protocols

gehaxelt:
Hi,

another PR of an useful feature (at least for me): A cmdline switch for the URL protocol (dir mode). 

I usually run gobuster on multiple domains in a for loop. This makes it more convenient, because I don't have to fiddle around and manually build the correct `-u` parameter. 

I'm not sure about the `-pr` switch name, though.

Sample runs:

```
$> ./gobuster  -m dir -w wordlist -u gehaxelt.in 

Gobuster v1.2                OJ Reeves (@TheColonial)
[+] Mode         : dir
[+] Url/Domain   : http://gehaxelt.in/
[+] Threads      : 10
[+] Wordlist     : wordlist
/.git/HEAD (Status: 301)
```

and 

```
$> ./gobuster  -m dir -w wordlist -u gehaxelt.in -pr https:// 

Gobuster v1.2                OJ Reeves (@TheColonial)
[+] Mode         : dir
[+] Url/Domain   : https://gehaxelt.in/
[+] Threads      : 10
[+] Wordlist     : wordlist
/.git/HEAD (Status: 301)
```

Kind regards

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,7 +15,6 @@ package main
 // Please see LICENSE file for license details.
 //
 //----------------------------------------------------
-
 import (
 	"bufio"
 	"crypto/tls"
@@ -71,6 +70,7 @@ type State struct {
 	Mode           string
 	NoStatus       bool
 	Password       string
+	Protocol       string
 	Printer        PrintResultFunc
 	Processor      ProcessorFunc
 	ProxyUrl       *url.URL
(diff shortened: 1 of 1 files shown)


OJ:
Thanks @gehaxelt, but this PR isn't necessary.

By default, if you use `dir` mode, `http` is used if a scheme isn't passed in when specifying the host. eg.
```
$ gobuster -u foo.com
```
is the same as:
```
$ gobuster -u http://foo.com
```
If you want to use https, just do:
```
$ gobuster -u https://foo.com
```

I think the `-pr` switch isn't necessary. If it's ok with you, I'll close this PR, as the existing functionality works fine.

PR #33, opened 2016-10-11
Added vendoring

insp3ctre:
The source is now self-contained, and does not have to rely on external dependencies.

Vendoring in Go: https://docs.google.com/document/d/1Bz5-UB7g2uPBdOx-rw5t9MxJwkfpx90cqG9AFL0JAYo/edit

If you prefer, you can use a dependency management tool like [Godep](https://github.com/tools/godep). This would make updating and managing your dependencies much easier, but the trade-off is that you would be relying on an unofficial tool that is not universally used (there is currently no "official" Go dependency management tool).

### DIFF
--- a/vendor/github.com/satori/go.uuid/.travis.yml
+++ b/vendor/github.com/satori/go.uuid/.travis.yml
@@ -1,0 +1,21 @@
+language: go
+sudo: false
+go:
+    - 1.2
+    - 1.3
+    - 1.4
+    - 1.5
+    - 1.6
+    - 1.7
+    - tip
+matrix:
+    allow_failures:
+        - go: tip
+    fast_finish: true
+before_install:
+    - go get github.com/mattn/goveralls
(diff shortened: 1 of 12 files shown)


OJ:
Thanks @insp3ctre I appreciate your efforts. My aim for this project was to just keep things really clean and simple, and while `go get` might be painful for some, I don't think it's a huge deal if people want to have a copy of it working locally. Also, given it's in Kali, most people are already covered for builds.

I'll ponder this a little more though. Thank you!

OJ:
Thanks again for this PR mate. I'm going to put this on ice as I think it's not necessary. I don't want you to think I'm ungrateful for your efforts though!

I'll close the PR, but it'll be here for future reference if the needs of the project change. Thank you again.

insp3ctre:
No worries, thanks for getting back to  me. Good luck with the project, I think it has some great potential. Cheers.

PR #44, opened 2017-05-13
Add feature to get HTTP OPTIONS

brimstone:
This adds a `-opt` feature to reveal HTTP OPTIONS.
I needed to be able to find dav shares for a CTF.
I didn't find any other tool that would do this, so I modified gobuster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ Yes, you're probably correct. Feel free to:
 * `-k` - Skip verification of SSL certificates.
 * `-l` - show the length of the response.
 * `-n` - "no status" mode, disables the output of the result's status code.
+* `-methods` Show HTTP methods support by the server at that URL. Good for exposing WebDAV support.
 * `-o <file>` - specify a file name to write the output to.
 * `-p <proxy url>` - specify a proxy to use for all requests (scheme much match the URL scheme).
 * `-r` - follow redirects.
--- a/libgobuster/dir.go
+++ b/libgobuster/dir.go
@@ -45,11 +45,11 @@ func (rh *RedirectHandler) RoundTrip(req *http.Request) (resp *http.Response, er
 }
 
 // Make a request to the given URL.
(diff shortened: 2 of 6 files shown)


OJ:
Thanks @brimstone. So is the intent here to just be able to discover the options available? Would it be worth adding filters here so that people can specify which options they're looking for in particular?

brimstone:
Correct, the intent is to use the HTTP method OPTIONS to discover the HTTP methods available to each directory or file discovered. Now that I type this out, maybe the flag should be `-methods`? There's <20 different possible http methods possible, and in practice there's <5 that keep showing up. I don't know that it makes sense to filter only matching methods, but maybe filter out the common: HEAD, GET, POST, OPTIONS methods. Another route could be a more explicit `-dav` option that determines if webdav is supported based on the methods returned by HTTP OPTIONS. What do you think?

OJ:
Sorry for not getting back to you earlier. I think it'd be great to be able to pass in a CSV of methods. From there people can use custom ones as well (given that it's possible to have custom methods in various services). The `-dav` option sounds nice but I say let's keep it simple to start with.

OJ:
Hi @brimstone sorry for the tardiness. I've landed a chunky refactor to the code, so would it be possible for you to update this PR so it's rebased on master? Thank you!

brimstone:
I'm looking into this now.

brimstone:
All I've done is rework this to account for the refactor to 1.4. This is still simply showing the server methods supported for a URL and not filtering them with just whay the user is requesting. I can add that if it's still desired, but it'll make the code longer.

0xdevalias:
Potentially related, may be relevant to consider with regards to this PR:
* https://github.com/OJ/gobuster/pull/76

OJ:
Apologies for the delay on this, can I please request that you rebase your changes on 2.0.0 now that I've finally landed #79? Thank you so much.

PR #47, opened 2017-05-28
Add ability to ignore responses within a size range

OJ:
This isn't ready for prime time yet, but I'm submitting the PR so I know I have to include it on the next milestone.

This PR will add support for filtering the responses based on file size. This means that if you are getting 200s that contain "Sorry, that file is not found" then you can use the size filter to ignore those responses. You can also do an inverse, and say you're only interested in responses that are in a certain size range.

### DIFF
--- a/main.go
+++ b/main.go
@@ -36,6 +36,11 @@ import (
 	"unicode/utf8"
 )
 
+const MaxUint = ^uint(0)
+const MinUint = 0
+const MaxInt = int(MaxUint >> 1)
+const MinInt = -MaxInt - 1
+
 // A single result which comes from an individual web
 // request.
 type Result struct {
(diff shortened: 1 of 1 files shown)


codingo:
Would love to see this land - this would also have value when trying to remove wildcard responses. If you could continue on a wildcard case (instead of stopping as it does now), then you could filter out responses between a range for a the wildcard, which you could identify by sending some guids in burp.

OJ:
I intend on getting on top of these puppies hopefully next week.

dsolstad:
What is the status on this?

OJ:
I'm keen to get this sorted in a more general sense. However, this has diverged from the current codebase so I'm going to have to rework it.  Watch this space, I'm having a break from work soon so I'll have more time to get through the PR backlog and implement some of these features.

IvoPereira:
@OJ is there any ETA for such a feature?

PR #42, opened 2017-03-12
Structure project in the boltdb-inspired-way

ilyaglow:
## Why
Attempt to address #41

## Summary
Main parts of gobuster package that I picked:
- dir.go
- dns.go
- helpers.go
- output.go
- parsecmd.go
- state.go

## How to try this out
```
git fetch origin pull/42/head:reorganize-main
git checkout reorganize-main
cd cmd/gobuster
go run main.go -h
```

## Links
[Boltdb reference](https://github.com/boltdb/bolt)

Any suggestions are welcomed

PR #51, opened 2017-08-18
Separate gobuster program/command line from core functionality

alias1:
There is an idea i'm playing with where it would be useful to be able to use the core of gobuster from another program, though because it is currently all a single 'program', there is no way to import it.

This should separate the command-line parsing/'program'  aspect of it from the core engine/functionality, which I've called `libgobuster` out of laziness (naming is hard..)

Should still be able to compile with `go get..` as before, and I don't believe it should make any functional difference to how anything runs; still compiles to a single binary, etc.

### DIFF
--- a/libgobuster/libgobuster.go
+++ b/libgobuster/libgobuster.go
@@ -1,0 +1,621 @@
+package libgobuster
+
+//----------------------------------------------------
+// Gobuster -- by OJ Reeves
+//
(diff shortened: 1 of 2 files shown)


alias1:
Just saw https://github.com/OJ/gobuster/pull/42, which might make this one obsolete

PR #54, opened 2017-08-30
Run gobuster from code

alias1:
This is branched off https://github.com/OJ/gobuster/pull/51, and further extends the 'libgobuster' concept by separating the program `State` setup from the command line argument parsing. While I feel like the state stuff could be cleaned up further, this basically allows another program to do something similar to the following:

```
import (
	"fmt"
	"os"
	"github.com/hashicorp/go-multierror"
	"github.com/OJ/gobuster/libgobuster"
)

func ConfigureState() (*libgobuster.State, *multierror.Error) {
	var extensions string
	var codes string
	var proxy string

	s := libgobuster.InitState()

	s.Threads = 10
	s.Mode = "dir"
	s.Wordlist = "/Users/devalias/sec/seclists/Discovery/Web_Content/common.txt"
	s.Url = "http://www.example.com/"

	extensions = ""
	codes = "200,204,301,302,307"
	proxy = ""

	if err := libgobuster.ValidateState(&s, extensions, codes, proxy); err.ErrorOrNil() != nil {
		return nil, err
	} else {
		return &s, nil
	}
}

func main() {
	state, err := ConfigureState()
	if err.ErrorOrNil() != nil {
		fmt.Printf("%s\n", err.Error())
		os.Exit(1)
	}

	libgobuster.Process(state)
}
```

In doing so, I converted a lot of the previously 'printed errors' in the state setup to use https://github.com/hashicorp/go-multierror instead. It should work exactly (or very close to it) the same for end users, but improves the programmatic usability. Eventually i'd like to rework things further to be able to get results back in a better way as well. I'll make those changes on this branch to simplify things if it isn't merged beforehand.

If https://github.com/OJ/gobuster/pull/42 is used (instead of #51) then a bunch of this work will need to be rebased/figured out how to merge on top of those changes (which I can do if/when the time comes).

This branch also includes formatting changes based on [gofmt](https://golang.org/cmd/gofmt/). Ideally, if we are going to do any merging between #42 and this, it would be good to run `gofmt` over it first, to minimize any potential merge errors from spacing changes.

### DIFF
--- a/libgobuster/libgobuster.go
+++ b/libgobuster/libgobuster.go
@@ -224,8 +224,6 @@ func GoGet(s *State, url, uri, cookie string) (*int, *int64) {
 	return MakeRequest(s, url+uri, cookie)
 }
 
-
-
 // Process the busting of the website with the given
 // set of settings from the command line.
 func Process(s *State) {
--- a/libgobuster/libgobuster.go
+++ b/libgobuster/libgobuster.go
@@ -1,0 +1,621 @@
+package libgobuster
+
+//----------------------------------------------------
+// Gobuster -- by OJ Reeves
+//
+// A crap attempt at building something that resembles
+// dirbuster or dirb using Go. The goal was to build
+// a tool that would help learn Go and to actually do
+// something useful. The idea of having this compile
+// to native code is also appealing.
+//
+// Run: gobuster -h
+//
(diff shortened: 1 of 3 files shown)


alias1:
Closing since merged into 1.4-dev

PR #56, opened 2017-09-19
Added custom prefix, suffix functionalities and -random-agent.

BBerastegui:
I added the option to specify custom prefix and suffix to the wordlist entries, just in case we want to try any kind of weird path combination, such as `%2f` or `//`, etc.


BBerastegui:
The -- seems not to be "oficially" supported, but it works if being use in the command line.

BBerastegui:
I also added the -random-agent feature with the user-agent dictionary from sqlmap.

OJ:
Hi @BBerastegui Thanks very much for this. I've just landed 1.4 (finally), sorry for the delay. Could you please rebase this against those changes? Thank you very much!

BBerastegui:
I'm trying to do so :P

But:

`libgobuster/dir.go:128: undefined: uuid.Must`

I can't find what's going on here. Any clue what I'm missing?

:'(

OJ:
Be sure to update your uuid lib code, that'll remove the compiler error.

BBerastegui:
As a reference, for those getting:

`undefined: uuid.Must` error when doing `go build .`

You should force update of the dependencies (doing only go get in the uuid library wasn't working for me):

`go get -u -v`

BBerastegui:
Me using Github like:



I'm sorry for this mess. Shame.

libgobuster/dir.go
```diff
@@ -59,6 +61,22 @@ func MakeRequest(s *State, fullUrl, cookie string) (*int, *int64) {
 		req.Header.Set("User-Agent", s.UserAgent)
 	}
 
+	if s.RandomUserAgent && s.UserAgent == "" {
+		b, err := ioutil.ReadFile("user-agents.txt")
+		if err != nil {
+			fmt.Print(err)
```
0xdevalias: You catch the error here, print it, but then execution will just continue I believe? We would probably be better returning the error here, or (maybe, but it's gross) panic'ing?

libgobuster/dir.go
```diff
@@ -59,6 +61,22 @@ func MakeRequest(s *State, fullUrl, cookie string) (*int, *int64) {
 		req.Header.Set("User-Agent", s.UserAgent)
 	}
 
+	if s.RandomUserAgent && s.UserAgent == "" {
+		b, err := ioutil.ReadFile("user-agents.txt")
```
0xdevalias: Hardcoded filename, is there a way to specify this from the CLI parameters?

libgobuster/state.go
```diff
@@ -36,39 +36,42 @@ type StringSet struct {
 // Contains State that are read in from the command
 // line when the program is invoked.
 type State struct {
-	Client         *http.Client
```
0xdevalias: Not sure why this is all marked as such a big block of changes. Did the whitespacing used change or something? Does this match `gofmt`?

BBerastegui: This does match gofmt, but I have no idea either.

libgobuster/dir.go
BBerastegui: How do you suggest to return if error then?

`return nil, nil` ?

0xdevalias: Normally it would be ‘return nil, er’ or similar. On my phone so can’t check the return type, may need to add a 3rd type (error) if it’s already returning 2. Then return nil, nil, error.

Can’t remember if I started implementing multierror in this or not, if so, possibly using that type might be beneficial.

Would also have to add an error check around anywhere that calls this.

Ideally we would be bubbling any errors back up to the highest point, so it can decide how to handle them (or die). Rather than the current panics we have scattered around the place.

0xdevalias: A little divergent from initial PR, but implementing that pattern for this function would help to clear up some of the existing tech debt.

libgobuster/state.go
0xdevalias: Haha, fair enough. Not a big issues

libgobuster/dir.go
BBerastegui: Function MakeRequest returns `(*int, *int64)` so it gives an error if you try to `return nil, err`.

I'll leave it as the rest of the returns in that function, as `nil, nil`

0xdevalias:
I would be making the return type (*int, *int64, error) then, and returning nil, nil, err. *as alluded to above*

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

BBerastegui:
GoGet must now then return one additional argument too:

`func GoGet(s *State, url, uri, cookie string) (*int, *int64, error) {
    return MakeRequest(s, url+uri, cookie)
}`

Would you like to keep sending this error up to the moon?

Before:
`wildcardResp, _ := GoGet(s, s.Url, fmt.Sprintf("%s", guid), s.Cookies)`
After:
`wildcardResp, _, err := GoGet(s, s.Url, fmt.Sprintf("%s", guid), s.Cookies)`

And same in all cases where GoGet is being called?

0xdevalias:
That was my intention, pushing error handling out to the ‘highest’ point where it makes logical sense to be handled. Which was also why I said it might be beyond the scope of this PR depending how much would have to change.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

BBerastegui:
I'll left `TODO`'s for error handling as you guys prefer to do it.

0xdevalias:
Thanks for making those changes @BBerastegui !

@OJ I haven't actually run this PR to confirm it all works as expected, but assuming it does, i'm pretty happy with how it looks from a code perspective. So if it works/you're happy, LGTM!

libgobuster/dir.go
```diff
@@ -59,6 +61,22 @@ func MakeRequest(s *State, fullUrl, cookie string) (*int, *int64) {
 		req.Header.Set("User-Agent", s.UserAgent)
 	}
 
+	if s.RandomUserAgent && s.UserAgent == "" {
+		b, err := ioutil.ReadFile(s.UserAgentsFile)
```
0xdevalias: We could probably make this more efficient by reading in the UserAgents file once at the beginning and caching it, rather than on every request. The current way might slow things down a bunch, particularly if there are a lot of agents in the file.

0xdevalias: Also.. is s.RandomUserAgent set from the parameters? At a quick skim of the 'changed files' here on GH, I can't see that parameter. Only the prefix/suffix.

README.md
```diff
@@ -53,8 +53,10 @@ Yes, you're probably correct. Feel free to:
 * `-n` - "no status" mode, disables the output of the result's status code.
 * `-o <file>` - specify a file name to write the output to.
 * `-p <proxy url>` - specify a proxy to use for all requests (scheme much match the URL scheme).
+* `-prefix <prefix>` - append a custom prefix for directory brute forces
 * `-r` - follow redirects.
 * `-s <status codes>` - comma-separated set of the list of status codes to be deemed a "positive" (default: `200,204,301,302,307`).
+* `-suffix <suffix>` - append a custom suffix for directory brute forces.
```
0xdevalias: Can you add the RandomUserAgent stuff to the README params as well?

libgobuster/dir.go
```diff
@@ -59,6 +61,22 @@ func MakeRequest(s *State, fullUrl, cookie string) (*int, *int64) {
...
+		if err != nil {
+			return nil, nil, err
+		}
+		lines := strings.Split(string(b), "\n")
+		var uas []string
+		for _, line := range lines {
+			if !strings.HasPrefix(line, "#") && len(line) > 0 {
```
0xdevalias: Not 100%, but you may need to trim pre/post whitespace on here, otherwise a blank line with a few spaces might be counted as a valid UA.

OJ:
Thanks for the efforts you're both putting into this :+1:

libgobuster/dir.go
BBerastegui: That's true. Where should I put it then?

I thought in SetupDir(), by adding `UserAgents    []string` to the State and using it to store the array of User-Agents.

But in this case, what do we do with the error? Panic? (as I put in the example), and also, this can't be done there (in SetupDir):

`        s.Printer = PrintDirResult
        s.Processor = ProcessDirEntry
        s.Setup = SetupDir`

As the ProcessDirEntry is run first in `statehelpers.go` and it already requires the User-Agent to be set.

0xdevalias: I think some of the checks belong in ValidateState() or ValidateDirModeState() if they aren't already: https://github.com/OJ/gobuster/blob/master/libgobuster/statehelpers.go

Adding `UserAgents []string` to the state would be the correct choice I think.

The 'easy' choice is to panic here, the 'better' choice is probably to add support for an 'error' return value to `SetupDir` as well. Which would also require similar to be done for `SetupDns`. And then handled somehow wherever they are called from:
* Added to the definition of `SetupFunc`  here: https://github.com/OJ/gobuster/blob/master/libgobuster/state.go#L24
* The error handling changed from checking a bool to checking for error here. Not actually sure what the bool is representing here.. we may need both it and the error..? But I think it's just meant to be a 'rudimentary' error check: https://github.com/OJ/gobuster/blob/master/libgobuster/state.go#L80
* There is a slight chance there are other changes required, these were the only ones I saw when looking just now though.

Bonus points if you then add an error return to Process: https://github.com/OJ/gobuster/blob/master/libgobuster/state.go#L76

And the associated error check for that at https://github.com/OJ/gobuster/blob/master/main.go#L76

Is `ProcessDirEntry`  actually run in `statehelpers`? The only line I see is https://github.com/OJ/gobuster/blob/master/libgobuster/statehelpers.go#L53 which I believe is just storing the pointer to the function. It's called elsewhere. So I think this would be fine?

libgobuster/dir.go
```diff
@@ -159,7 +148,7 @@ func ProcessDirEntry(s *State, word string, resultChan chan<- Result) {
 	// TODO: Error propagation handling
 	dirResp, dirSize, err := GoGet(s, s.Url, prefix+word+suffix, s.Cookies)
 	if err != nil {
-		panic(err)
+		//panic(err)
```
BBerastegui: I made this terrible thing here because I realized that the 302 redirect was causing a panic here.

0xdevalias: Which '302 redirect'? Can you give a specific example/site it happens on for debugging.

Definitely will need to actually do something with the error here before it can be merged, otherwise it will fail silently, and that's bad.

Without running it/seeing it in practice I can't really give a good idea of how it should be handled. If it's returning an error where it isn't really an error, then we can probably catch that case and handle it somewhere appropriately.

libgobuster/statehelpers.go
```diff
@@ -170,6 +171,22 @@ func ValidateDirModeState(
 		}
 	}
 
+	// Read the user agents file
+	if errorList.ErrorOrNil() == nil && s.RandomUserAgent {
+		b, err := ioutil.ReadFile(s.UserAgentsFile)
+		if err != nil {
+			errorList = multierror.Append(errorList, fmt.Errorf("[!] Error when reading the User-Agents file"))
```
0xdevalias: We're masking the cause of the error here by not including it. Could just add the error directly to the errorList, though then you can't add your additional context. At the very least I would include the error message with something like:

```
fmt.Errorf("[!] Error when reading the User-Agents file: %s", err)
```

Even that isn't super ideal. Could potentially use `%+v` to get more context in it? In my own code, I normally would wrap it with [another library](https://github.com/pkg/errors):

```
errors.Wrap(err, "[!] Error when reading the User-Agents file")
```

That would be my best case scenario to do things 'right', but would require adding in another dependency (which personally I don't see as a huge issue).

libgobuster/statehelpers.go
```diff
@@ -170,6 +171,22 @@ func ValidateDirModeState(
...
+		}
+		lines := strings.Split(string(b), "\n")
+		var uas []string
+		for _, line := range lines {
+			if !strings.HasPrefix(line, "#") && len(line) > 0 {
```
0xdevalias: I would probably call `strings.TrimSpace` on the line before checking for the prefix:

* https://golang.org/pkg/strings/#TrimSpace

libgobuster/statehelpers.go
```diff
@@ -170,6 +171,22 @@ func ValidateDirModeState(
 		}
 	}
 
+	// Read the user agents file
+	if errorList.ErrorOrNil() == nil && s.RandomUserAgent {
+		b, err := ioutil.ReadFile(s.UserAgentsFile)
```
0xdevalias: `ioutil.ReadFile` will read the entire file into memory, then you split on newlines, remove comments/etc. It may not matter super much here, but there is a potential that we read in a lot of stuff that is never required, etc.

Following the pattern used for reading the , we could use `bufio.NewScanner` to read/process the file line by line:

* https://github.com/OJ/gobuster/blob/master/libgobuster/state.go#L143
* https://github.com/OJ/gobuster/blob/master/libgobuster/state.go#L158-L168

So an implementation in this case might look something like:

```
for scanner.Scan() {
		if s.Terminate {
			break
		}
		line := strings.TrimSpace(scanner.Text())

		// Skip "comment" (starts with #), as well as empty lines
		if !strings.HasPrefix(word, "#") && len(line) > 0 {
			uas = append(uas, line)
		}
	}
```

If we wanted to be fancy, we could probably refactor the common functionality between them into a helper function.. But that is probably beyond what is required here.

libgobuster/dir.go
BBerastegui: Funny thing is that now, when trying to reproduce it, I realised that it's happening in the host I was testing it, for a 302 redirect:

```./gobuster -u https://king.com/ -w <(echo -n "images")

Gobuster v1.4.1              OJ Reeves (@TheColonial)
[+] Mode         : dir
[+] Url/Domain   : https://king.com/
[+] Threads      : 10
[+] Wordlist     : /dev/fd/63
panic: Redirect code: 302

goroutine 15 [running]:
panic(0x22a900, 0xc4203d1ea8)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
github.com/bberastegui/gobuster/libgobuster.ProcessDirEntry(0xc4200ce000, 0xc4203dc16a, 0x6, 0xc4203109c0)
	/[REDACTED]/go/src/github.com/bberastegui/gobuster/libgobuster/dir.go:151 +0x438
github.com/bberastegui/gobuster/libgobuster.Process.func1(0xc420310960, 0xc4200ce000, 0xc4203109c0, 0xc4203dc1c0)
	/[REDACTED]/go/src/github.com/bberastegui/gobuster/libgobuster/state.go:116 +0x8f
created by github.com/bberastegui/gobuster/libgobuster.Process
	/[REDACTED]/go/src/github.com/bberastegui/gobuster/libgobuster/state.go:122 +0x1a9```

But not on other redirects like 301, for example.

libgobuster/statehelpers.go
BBerastegui: I like the `%+v`:

```./gobuster -u https://google.com/ -w <(echo "images") -random-agent

Gobuster v1.4.1              OJ Reeves (@TheColonial)
1 error occurred:

* [!] Error when reading the User-Agents file: open user-agents.txt: no such file or directory```

Shaked:
Hey @BBerastegui do you need help with this pull request? Seems very useful!

BBerastegui:
Hey @Shaked, to be completely honest, I think my knowledge of Go is pretty limited and I think I can't catch up with this now.

Feel free to get my commit and rebase the changes agains the latest version if you want :)

OJ:
I'll see if I can put some time into this later. Sorry for the epic divergence. Dat Open Source Lyfe yo!

FireFart:
This should all be already implemented on the `v3.1-working` branch

PR #71, opened 2018-01-16
Add ability to bust with placeholders.

leonjza:
## proposal 

This commit adds the ability to define placeholder keywords in a URL and have those replaced with words defined by stdin or a wordlist.

An example use would be:

```
$ gobuster -w wordlist -u http://localhost:8000/BUSTbar.txt -ph BUST -v -e
Missed: http://localhost:8000/BUSTbar.txt/foo (Status: 404)
Missed: http://localhost:8000/BUSTbar.txt/bar (Status: 404)
Missed: http://localhost:8000/BUSTbar.txt/baz (Status: 404)
Found : http://localhost:8000/BUSTbar.txt/(replaced BUST with foo) (Status: 200)
Missed: http://localhost:8000/BUSTbar.txt/(replaced BUST with baz) (Status: 404)
Missed: http://localhost:8000/BUSTbar.txt/(replaced BUST with bar) (Status: 404)
```

A URL such as http://localhost:8000/BUSTbarBUST.txt will have the processor replace both instances of the BUST placeholder word.

## Problems:
Printing matched results is not great. Ideally one would like to see something like the following output: `Found: foobar.txt (Status: 200)`. Before I go and get a fresh bit of sticky tape and gum, is there a preferred way to handle this?
I also can't seem to find where the trailing slash is appended to the URL, even though I have not specified `-f`.

### DIFF
--- a/libgobuster/dir.go
+++ b/libgobuster/dir.go
@@ -136,6 +136,21 @@ func ProcessDirEntry(s *State, word string, resultChan chan<- Result) {
 		}
 	}
 
+	if s.Placeholder != "" {
+
+		replacedURL := strings.Replace(s.Url, s.Placeholder, word, -1)
+		replacedURL = strings.TrimRight(replacedURL, "/") // halp! where does the trailing / come from?
+
+		dirResp, dirSize := GoGet(s, replacedURL, "", s.Cookies)
+		if dirResp != nil {
+			resultChan <- Result{
+				Entity: "(replaced " + s.Placeholder + " with " + word + ")",
+				Status: *dirResp,
+				Size:   dirSize,
+			}
+		}
+	}
+
 	// Follow up with files using each ext.
 	for ext := range s.Extensions {
 		file := word + s.Extensions[ext]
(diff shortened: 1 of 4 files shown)


hypn:
I was just looking for this functionality - would love this to be merged in.

BBerastegui:
Isn't the same behaviour able to be reproduced with custom prefix and suffix?

Ref: #56

leonjza:
@BBerastegui I don't see the possibility to have multiple pre/suffixes with that PR? Happy to have that there if its possible ;)

BBerastegui:
It's true that it's not the same behaviour, my bad 👎

0xdevalias:
There is a nuanced difference between the 2, but I think both should be considered together, and figure what the best pattern is to support both in a 'clean' way.

OJ:
Any chance of the rebase happening on v3.0-working please @leonjza ? Sorry for the nudge, trying to get through the PRs now.

leonjza:
Sure let me give it a shot.

leonjza:
Closing this in favour of #132

PR #73, opened 2018-01-23
Added negative status code option

dave-au:
I've added "-S" to define negative status codes (instead of "-s").
There are no defaults for "-S".
If "-S" is specified, the values provided will overwrite the "-s" defaults (or any values specified by the user for "-s").

Example would be...
gobuster -m dir -w words -u http://host/ -S 404

*fingers crossed* This is literally my first pull request ever.  I hope it is OK.

### DIFF
--- a/libgobuster/dir.go
+++ b/libgobuster/dir.go
@@ -108,16 +108,28 @@ func SetupDir(s *State) bool {
 	guid := uuid.Must(uuid.NewV4())
 	wildcardResp, _ := GoGet(s, s.Url, fmt.Sprintf("%s", guid), s.Cookies)
 
-	if s.StatusCodes.Contains(*wildcardResp) {
-		s.IsWildcard = true
-		fmt.Println("[-] Wildcard response found:", fmt.Sprintf("%s%s", s.Url, guid), "=>", *wildcardResp)
-		if !s.WildcardForced {
-			fmt.Println("[-] To force processing of Wildcard responses, specify the '-fw' switch.")
+	if s.NegStatusCodes.Stringify() == "" {
+		if s.StatusCodes.Contains(*wildcardResp) {
+			s.IsWildcard = true
+			fmt.Println("[-] Wildcard response found:", fmt.Sprintf("%s%s", s.Url, guid), "=>", *wildcardResp)
+			if !s.WildcardForced {
+				fmt.Println("[-] To force processing of Wildcard responses, specify the '-fw' switch.")
+			}
+			return s.WildcardForced
 		}
-		return s.WildcardForced
+		return true
+	} else {
+		if !s.NegStatusCodes.Contains(*wildcardResp) {
+			s.IsWildcard = true
+			fmt.Println("[-] Wildcard response found:", fmt.Sprintf("%s%s", s.Url, guid), "=>", *wildcardResp)
+			if !s.WildcardForced {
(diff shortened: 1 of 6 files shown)


libgobuster/dir.go
```diff
@@ -108,16 +108,28 @@ func SetupDir(s *State) bool {
 	guid := uuid.Must(uuid.NewV4())
 	wildcardResp, _ := GoGet(s, s.Url, fmt.Sprintf("%s", guid), s.Cookies)
 
-	if s.StatusCodes.Contains(*wildcardResp) {
-		s.IsWildcard = true
-		fmt.Println("[-] Wildcard response found:", fmt.Sprintf("%s%s", s.Url, guid), "=>", *wildcardResp)
-		if !s.WildcardForced {
-			fmt.Println("[-] To force processing of Wildcard responses, specify the '-fw' switch.")
+	if s.NegStatusCodes.Stringify() == "" {
```
0xdevalias: It would probably be better to use `len`?

* https://godoc.org/golang.org/x/tools/container/intsets#Sparse.Len

libgobuster/dir.go
```diff
@@ -156,34 +168,68 @@ func PrintDirResult(s *State, r *Result) {
 
 	// Prefix if we're in verbose mode
 	if s.Verbose {
-		if s.StatusCodes.Contains(r.Status) {
-			output = "Found : "
+		if s.NegStatusCodes.Stringify() == "" {
```
0xdevalias: As above:

* https://godoc.org/golang.org/x/tools/container/intsets#Sparse.Len

libgobuster/helpers.go
```diff
@@ -59,7 +59,11 @@ func ShowConfig(s *State) {
 		}
 
 		if s.Mode == "dir" {
-			fmt.Printf("[+] Status codes : %s\n", s.StatusCodes.Stringify())
+			if s.NegStatusCodes.Stringify() == "" {
```
0xdevalias: As above:

* https://godoc.org/golang.org/x/tools/container/intsets#Sparse.Len

libgobuster/statehelpers.go
```diff
@@ -86,7 +88,7 @@ func ValidateState(
 	}
 
 	if s.Mode == "dir" {
-		if err := ValidateDirModeState(s, extensions, codes, proxy, errorList); err.ErrorOrNil() != nil {
+		if err := ValidateDirModeState(s, extensions, codes, negcodes, proxy, errorList); err.ErrorOrNil() != nil {
```
0xdevalias: Given they are mutually exclusive, can we just pass in 'codes' and then the state of the `-s` or `-S` (possibly passed as a bool somewhere) determines whether we treat them as 'positive' or 'negative' code? Feels cleaner.

OJ:
Hey @dave-au, you still keen for this? v3.0-working is our new working branch, would love to see you get your PR rebased onto that so we can move ahead. Cheers, and apologies for the delay.

cbaker730:
I hope @dave-au adds this - I was just looking for this option and would love to see it included

FireFart:
We now have this feature in v3:
```
 -s, --statuscodes string            Positive status codes (will be overwritten with statuscodesblacklist if set) (default "200,204,301,302,307,401,403")
  -b, --statuscodesblacklist string   Negative status codes (will override statuscodes if set)
```

PR #77, opened 2018-04-10
Feature: numeric range expansion

objectified:
I've added a feature to gobuster to support numeric range expansion in words. For example, if you have an entry in a wordlist like this:

someapp-[1-3].[12-15].[1-2]

gobuster can now expand that word into a number of words, to scan for all combinations of the numeric ranges specified. For instance, the above pattern would expand into:

```
    someapp-1.12.1

    someapp-2.12.1

    someapp-3.12.1

    someapp-1.13.1

    ...
```
... and so on.

It also works for URLs (e.g. /foo/bar-[1-3].[12-15]) and DNS enumeration (www[1-100] and so on).

I've introduced a new command line parameter '-xr' to explicitly enable this feature, as it might be a bit more CPU intensive.

I've also added a number of unit tests for the feature, which can be found in libgobuster/wordranges_test.go , and they can be executed as such:

``
    cd libgobuster && go test -v
``
I've also adjusted the README.md to describe the new parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Yes, you're probably correct. Feel free to:
 * `-u <url/domain>` - full URL (including scheme), or base domain name.
 * `-v` - verbose output (show all results).
 * `-w <wordlist>` - path to the wordlist used for brute forcing.
-* `-xr` - expand numeric ranges in wordlist (example:foo-[1-10]).
+* `-xr` - expand numeric ranges in wordlist (example: foo-[1-10]).
 
 ### Command line options for `dns` mode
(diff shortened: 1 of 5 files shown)


OJ:
Nifty feature, thanks @objectified. I'm trying to work through the backlog of PRs at the moment and I hope to get to this soon!

objectified:
Too much seems to have changed since the huge refactoring PR for this to rebase sanely. I'll rewrite the feature on top of the refactored codebase, and close this PR for now.

PR #79, opened 2018-05-14
implement status on dir mode

FireFart:
This implements a progress output in dir mode.

I also changed some golint related stuff (mostly formatting)

Example run:
```
root@kali:~# ./gobuster -u http://10.1.1.1/ -w /usr/share/wordlists/dirb/common.txt 

Gobuster v1.4.1              OJ Reeves (@TheColonial)
[+] Mode         : dir
[+] Url/Domain   : http://10.1.1.1/
[+] Threads      : 10
[+] Wordlist     : /usr/share/wordlists/dirb/common.txt
/index.html (Status: 200)
/robots.txt (Status: 200)
Status: 4613 / 4614
```

### DIFF
--- a/libgobuster/dir.go
+++ b/libgobuster/dir.go
@@ -1,6 +1,7 @@
 package libgobuster
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -101,16 +102,16 @@ func MakeRequest(s *State, fullUrl, cookie string) (*int, *int64) {
 // Small helper to combine URL with URI then make a
 // request to the generated location.
 func GoGet(s *State, url, uri, cookie string) (*int, *int64) {
-	return MakeRequest(s, url+uri, cookie)
+	return MakeRequest(s, fmt.Sprintf("%s%s", url, uri), cookie)
 }
 
 func SetupDir(s *State) bool {
 	guid := uuid.Must(uuid.NewV4())
-	wildcardResp, _ := GoGet(s, s.Url, fmt.Sprintf("%s", guid), s.Cookies)
+	wildcardResp, _ := GoGet(s, s.URL, fmt.Sprintf("%s", guid), s.Cookies)
 
 	if s.StatusCodes.Contains(*wildcardResp) {
 		s.IsWildcard = true
-		fmt.Println("[-] Wildcard response found:", fmt.Sprintf("%s%s", s.Url, guid), "=>", *wildcardResp)
+		fmt.Println("[-] Wildcard response found:", fmt.Sprintf("%s%s", s.URL, guid), "=>", *wildcardResp)
 		if !s.WildcardForced {
 			fmt.Println("[-] To force processing of Wildcard responses, specify the '-fw' switch.")
 		}
@@ -126,11 +127,15 @@ func ProcessDirEntry(s *State, word string, resultChan chan<- Result) {
 		suffix = "/"
 	}
 
+	s.Mu.Lock()
+	s.WordlistPosition++
+	s.Mu.Unlock()
+
 	// Try the DIR first
-	dirResp, dirSize := GoGet(s, s.Url, word+suffix, s.Cookies)
+	dirResp, dirSize := GoGet(s, s.URL, fmt.Sprintf("%s%s", word, suffix), s.Cookies)
 	if dirResp != nil {
 		resultChan <- Result{
-			Entity: word + suffix,
+			Entity: fmt.Sprintf("%s%s", word, suffix),
 			Status: *dirResp,
 			Size:   dirSize,
 		}
@@ -138,8 +143,8 @@ func ProcessDirEntry(s *State, word string, resultChan chan<- Result) {
 
 	// Follow up with files using each ext.
 	for ext := range s.Extensions {
-		file := word + s.Extensions[ext]
-		fileResp, fileSize := GoGet(s, s.Url, file, s.Cookies)
+		file := fmt.Sprintf("%s%s", word, s.Extensions[ext])
+		fileResp, fileSize := GoGet(s, s.URL, file, s.Cookies)
 
 		if fileResp != nil {
 			resultChan <- Result{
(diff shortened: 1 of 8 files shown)


eur0pa:
This branch fails to compile atm:

```
# _/Users/europa/pentest/gobuster-timeout
./main.go:42:19: s.URL undefined (type libgobuster.State has no field or method URL, but does have Url)
./main.go:49:21: s.Timeout undefined (type libgobuster.State has no field or method Timeout)
```

FireFart:
@eur0pa because you need to change `"github.com/OJ/gobuster/libgobuster"` in `main.go` to `"github.com/FireFart/gobuster/libgobuster"` if you are using my repo

eur0pa:
Tested this PR for a few days; it still hangs from time to time but, worst, it will still sometime quit inexplicably after a few hundred requests

FireFart:
I'm currently working on a big refactor so I hope this error is also fixed (stumbled across it also multiple times)

eur0pa:
Seems to be working well for now, except when using extensions: the status progress won't take them into account, as it only counts the number of lines in the wordlist and not the amount of requests-to-be

FireFart:
thx fixed it to now be a request counter instead of a line counter

eur0pa:
Serious improvement

FireFart:
Jeah found it during tracing. The error is also present in the current released gobuster so it creates a lot of connections to the target slowing things down. If I apply this patch to the current master I also get a ~3 times speedup

```patch
--- a/libgobuster/dir.go
+++ b/libgobuster/dir.go
@@ -2,6 +2,7 @@ package libgobuster

 import (
        "fmt"
+       "io"
        "io/ioutil"
        "net/http"
        "net/url"
@@ -93,6 +94,8 @@ func MakeRequest(s *State, fullUrl, cookie string) (*int, *int                                                                                                                                                                                                                                                                                           64) {
                } else {
                        *length = resp.ContentLength
                }
+       } else {
+               io.Copy(ioutil.Discard, resp.Body)
        }

        return &resp.StatusCode, length
```

OJ:
There are things about this that are awesome, but I'm not going to lie it's an epic monolith. Is there a chance we can pick some of this apart? The wordlist status thing would be perfect to pull out to make the overall refactor simpler.

FireFart:
@OJ I know :/ I think extracting things will be hard because there was a lot that changed internally for example for the status output that needs to be reimplemented on the current version and then changed again for the refactoring. What do you think about doing unit tests instead so we can verify the behaviour of single functions?

PS: The speedup patch is simple to apply to current master, just apply the diff from above

OJ:
Hey @FireFart I have it on my TODO list to have a good look at this today. Sorry for the delay!

Thanks so much for all your work so far.

FireFart:
@OJ awesome thanks! I started with first unit tests yesterday. If you need help just ping me!

anshumanbh:
@FireFart I believe this breaks because both `gobusterdir.go` and `gobusterdns.go` use `github.com/OJ/gobuster/libgobuster` as one of the imports and I believe that still uses the old uuid package because I see this while building the Docker image:

```
vendor/github.com/OJ/gobuster/libgobuster/dir.go:116:19: not enough arguments in call to uuid.Must
	have (uuid.UUID)
	want (uuid.UUID, error)
vendor/github.com/OJ/gobuster/libgobuster/dns.go:13:19: not enough arguments in call to uuid.Must
	have (uuid.UUID)
	want (uuid.UUID, error)
```

FireFart:
@anshumanbh it looks like your code uses the "real" libgobuster from current master (that's not going to work). I added some test instructions at the end of the PR description to get it up and running. Which Docker image are you trying to build?

anshumanbh:
@FireFart Doesn't the status branch of your forked repo also use the real libgobuster. `https://github.com/FireFart/gobuster/blob/status/gobusterdns/gobusterdns.go#L10`. The import statement here `github.com/OJ/gobuster/libgobuster` uses the master branch of gobuster, corret? Am I missing something?

I have a Dockerfile that looks like this:

```
# Build Container
FROM golang:1.9.4-alpine3.7 AS build-env
RUN apk add --no-cache --upgrade git openssh-client ca-certificates
RUN go get -u github.com/golang/dep/cmd/dep

WORKDIR /go/src/github.com/firefart/gobuster

# Cache the dependencies early
COPY Gopkg.toml Gopkg.lock ./
RUN dep ensure -vendor-only -v

COPY gobusterdir/ gobusterdir/
COPY gobusterdns/ gobusterdns/
COPY libgobuster/ libgobuster/
COPY main.go ./

# Install the binary
RUN go install

# Final Container
FROM alpine:3.7
COPY --from=build-env /go/bin/gobuster /usr/bin/gobuster

# Copy wordlists
COPY wordlists wordlists

ENTRYPOINT ["/usr/bin/gobuster"]
```

And, I am using `dep` to manage the dependencies.

eur0pa:
There seems to be a problem with the http library, no matter how you shuffle this code, it won't reuse connections:

```
~ netstat -antup|grep -c TIME_WAIT
(Not all processes could be identified, non-owned process info
 will not be shown, you would have to be root to see it all.)
402
```

This was with 50 threads. I tried implementing a simpler bruteforcer in go and still go the same behavior, no matter how many tricks I try:

```
	tr = &http.Transport{
		MaxIdleConnsPerHost: *threads,
		MaxIdleConns:        100,
		IdleConnTimeout: time.Duration(*timeout) * time.Second,
		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
	}
```

Also reading / discarding body doesn't change this behavior, neither in my tool nor gobuster. Any ideas?

FireFart:
@eur0pa I think TIME_WAIT is not directly related to connection reuse, it's a socket thing. Try watching the requests in Wireshark and have a look at the source ports from your machine. If they stay the same, connections are reused. That's how I checked it after implementing it.

eur0pa:
@FireFart yeah I went the sysctl way to solve TIME_WAIT both for gobuster and other tools. What really boggles my mind though, is how bad the Go resolver really is. Even in gobuster, any request will also open a socket to to a random resolver picked from `/etc/resolv.conf`. After a certain amount of _time_, even with all sockets closing down properly, tools will fail with `lookup [DOMAIN] on [DNS]:53: no such host`: this will cause requests to fail and files to be missed, a bunch at a time. 
In order to reproduce this quickly, serve a large list of hosts to gobuster via parallel and see what happens, ie:

```
[ assuming hosts.txt has, say, 200 entries ]
~ cat hosts.txt | parallel --bar -j 20 "gobuster -t 10 -w wordlist.txt -u {}"
```

You'll see that everything will start to fail pretty soon. Even by lowering the jobs amount down to 1 or 2 will eventually see it fail. This doesn't happen in Python, and seems to be a long-standing issue with Go (https://github.com/golang/go/issues/18588). Really no idea how to solve this, even in gobuster.

joohoi:
What's the current state of this PR in general? I'm thinking of starting to write a new PR, but unsure if I should base it on the refactored code in this branch, or the current master. Sorry for the static!

OJ:
The status is that this PR is a monster that includes a lot more than a
code refactor. It's taking a lot of time for me to go through it. Time is
something in particularly short on at the moment.

I would hold off submitting anything until this is out of the way. Sorry
mate.

PR #89, opened 2018-07-14
Added ability to search for files only

rtheory:
This minor change adds the -xo switch to enable file-only searching. This prevents gobuster from searching directory names first. This is useful when you've already performed a directory search earlier and now want to search for files by extension without re-querying for directories again. This improves speed and removes unnecessary queries.

* Added -xo switch processing
* Added -xo help statement in -h
* Added error handling; -xo can only be set if -x is provided
* Added line to gobuster output:  [+] Files Only   : true

This could be done other ways, but ultimately I chose just making "-xo" since it fits right next to the companion "-x" that everyone is already familiar with.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ Flags:
   -c, --cookies string                Cookies to use for the requests
   -e, --expanded                      Expanded mode, print full URLs
   -x, --extensions string             File extension(s) to search for
+  -X, --filesonly                     Search only for files in a known directory, skips directory search
   -r, --followredirect                Follow redirects
   -H, --headers stringArray           Specify HTTP headers, -H 'Header1: val1' -H 'Header2: val2'
   -h, --help                          help for dir
--- a/cli/cmd/dir.go
+++ b/cli/cmd/dir.go
@@ -119,6 +119,11 @@ func parseDirOptions() (*libgobuster.Options, *gobusterdir.OptionsDir, error) {
 		return nil, nil, fmt.Errorf("invalid value for wildcard: %v", err)
 	}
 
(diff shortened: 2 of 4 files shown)


libgobuster/dir.go
```diff
@@ -134,13 +134,15 @@ func ProcessDirEntry(s *State, word string, resultChan chan<- Result) {
...
 	}
 
-	// Try the DIR first
-	dirResp, dirSize := GoGet(s, s.Url, word+suffix, s.Cookies)
-	if dirResp != nil {
-		resultChan <- Result{
-			Entity: word + suffix,
-			Status: *dirResp,
-			Size:   dirSize,
+	if !(s.ExtensionsOnly) {
```
OJ: The use of parentheses here isn't necessary.

OJ:
Ping @rtheory would you mind rebasing on the new master now that 3.0 has been released? Many thanks!

rtheory:
Wow, a lot has changed since I originally wrote this! I rebased to the new master and re-implemented the option based on the new ways gobuster works. Testing worked well:

`gobuster dir -x txt,docx -w test.txt -u https://example.com`
This is standard usage. It makes requests for directories and file names as expected.

`gobuster dir -x txt,docx -X -w test.txt -u https://example.com`
Same as above, but there are no requests for directories, only file names

libgobuster/dir.go
rtheory: Good point. Fixed in the new file that this is now associated with: gobusterdir.go

OJ:
Great, thank you for your efforts and support @rtheory, I'll look to get this merged pronto! We'll do a smaller release for 3.0.1 when this and a couple of other small things are ready.

PR #92, opened 2018-08-29
fix increment error

### LINKED ISSUE #91: progress bar is not working in new version
Progress bar isnt moving during the wordlist run

```
$ ./gobuster -w wordlist/cg_big_wordlist.txt -u yahoo.com -m dns -t 10 -i
Gobuster v2.0.0              OJ Reeves (@TheColonial)
[+] Mode         : dns
[+] Url/Domain   : yahoo.com
[+] Threads      : 10
[+] Wordlist     : wordlist/cg_big_wordlist.txt
2018/08/29 21:09:38 Starting gobuster
Found: 360.yahoo.com [98.136.96.140, 98.136.101.175, 2001:4998:44:3204::1000, 2001:4998:24:705::3000]
Found: 3d.yahoo.com [98.136.103.23]
Progress: 0 / 97371 (0.00%)
```

let the whole thing run and stayed at progress 0 :-(

FireFart:
This fixes #91.

The progress counter is now a word counter and no more request counter anymore. libgobuster can't really predict how many requests a plugin will make so this is the only sane way of doing it without putting the plugin in charge of updating the counter

### DIFF
--- a/libgobuster/libgobuster.go
+++ b/libgobuster/libgobuster.go
@@ -111,7 +111,6 @@ func (g *Gobuster) ClearProgress() {
 // GetRequest issues a GET request to the target and returns
 // the status code, length and an error
 func (g *Gobuster) GetRequest(url string) (*int, *int64, error) {
-	g.incrementRequests()
 	return g.http.makeRequest(url, g.Opts.Cookies)
 }
 
@@ -126,6 +125,7 @@ func (g *Gobuster) worker(wordChan <-chan string, wg *sync.WaitGroup) {
 			if !ok {
 				return
 			}
+			g.incrementRequests()
 			// Mode-specific processing
 			res, err := g.plugin.Process(g, word)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


FireFart:
oh this causes errors on urls + extensions

libgobuster/libgobuster.go
```diff
@@ -111,10 +112,19 @@ func (g *Gobuster) ClearProgress() {
...
+// DNSLookup looks up a domain via system default DNS servers
+func (g *Gobuster) DNSLookup(domain string) ([]string, error) {
+	return net.LookupHost(domain)
+}
+
+// DNSLookupCname looks up a CNAME record via system default DNS servers
+func (g *Gobuster) DNSLookupCname(domain string) (string, error) {
+	return net.LookupCNAME(domain)
+}
+
```
OJ: I think I'd prefer to see both of these things go down into the individual plugins. Right now the plugins call out to these three functions (`GetRequest`, `DNSLookup` and `DNSLookupCname`), yet those functions either call default Golang libs or call back into the plugin. Are you open to change this so that these functions disappear?

OJ: Nevermind. Let's do this as part of a different piece of work.

PR #94, opened 2018-08-31
nuke test dep

FireFart:
This removes the test dependency and uses go core functionality instead

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,5 @@ go:
   - "1.8"
   - "1.10.x"
   - master
+
+script: make test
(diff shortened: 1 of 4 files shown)


FireFart:
After this has been merged we can inform upstream kali via their bugtracker (https://bugs.kali.org/view.php?id=4939) that their patch to remove the test file because of the missing dependency is not needed any more (http://git.kali.org/gitweb/?p=packages/gobuster.git;a=tree;f=debian/patches;h=3fc2625ca44144e1f849013d1a524d2ec8d33fb1;hb=refs/heads/kali/master) cc @OJ

OJ:
Do we need to do a full updated version (ie. 2.0.2) for the changes to be picked up?

FireFart:
Good question. We could just open a bug on the bugtracker and ask if they need a full version. Also they should use the makefile for building as it now shrinks the binaries.

PR #107, opened 2018-10-09
implement vhost based bruteforcing

FireFart:
This PR implements simple VHOST based bruteforcing. It sends requests with a modified `Host` header and checks if the specified vhost is available.

I moved the HTTP stuff back into `libgobuster` because now a second plugin needs http functionality

Follow ups after this one: https://github.com/OJ/gobuster/pull/137 and https://github.com/OJ/gobuster/pull/138


OJ:
Can't help but notice we have a fair bit of code duplication here when compared to `dir` mode. Is there a way to reduce this somehow? Given they're both effectively HTTP brute forcers, can we not put common elements somewhere that can be shared?

FireFart:
@OJ I merged the v3.0 working changes into this branch so it's now a clean PR.

Which code do you exactly mean? The new `Run` method is pretty short and the `GetConfigString` method is based on the cobra parameters (so I would leave it in this file completely). The only thing we could extract is the `PreRun` method but I don't think it makes sense for these around 20 lines

FireFart:
Or do you mean the cobra stuff itself with the shared parameters?

cli/cmd/vhost.go
```diff
@@ -0,0 +1,178 @@
...
+				return nil, nil, fmt.Errorf("url scheme not specified")
+			} else if port == 80 {
+				plugin.URL = fmt.Sprintf("http://%s", plugin.URL)
+			} else {
+				plugin.URL = fmt.Sprintf("https://%s", plugin.URL)
+			}
+		}
+	}
+
+	cookies, err := cmdVhost.Flags().GetString("cookies")
```
OJ: Probably no need to have temp variables every time we parse a flag. I think it makes more sense to write them straight to the plugin content instead. We throw errors if they're invalid anyway.
```go
plugin.Cookies, err = cmdVhost.Flags().GetString("cookies")
```
Same goes for all the options parsed below.

OJ:
I was about to pull the trigger on merging this, but then I head a thought.

This caters for bruteforcing vhost names when the base domain name is known. However, what if a webserver supports multiple domain names? So a single host with IP x.x.x.x might serve bar.com, foo.bar.com, foo.com and bar.foo.com.

Do we want to be able to support that or are we going to stick with a single base domain name?

I also thought "Hang on, isn't this just DNS bruteforcing via HTTP", and while I think that for the most part this is how it'll be used (so DNS is a better option), there will be cases where external DNS entries might not be available, yet are still supported. It's an edge case, for sure, but probably still worth having.

OJ:
Pardon the force-push. I rebased onto v3.0 working after adding some readme stuff to the base.

FireFart:
I use it most of the time on some CTF style boxes like on hackthebox to find VHOSTs in named vhost based setups. On these scenarios you can rarely enumerate a DNS server because they are fake hostnames not existing.

But nice to see some action on this PR again 👍

OJ:
Yeah cool, that's exactly the kind of scenario I pictured. All good.

So what do you think with regards to the base domain thing? Do you think that extends the scope too much? I'm happy to go with what we have as it's a handy feature, and changes are that servers aren't necessarily going to support lots of subdomains across different domains unless they wildcard stuff.

Yeah sorry it took so long, been SOO busy :)

FireFart:
I think it's good for now to focus on one domain and bruteforce it. We can still think of adding a feature to also enumerate the base domain in the future

OJ:
Hey @FireFart so the one thing that seems to be left outstanding for me is that the vhost bruteforcing doesn't track status codes other than `200`. If a `301` is found, shouldn't that mean that something was discovered?

Not sure how to handle this case. Is vhost brute forcing only about finding different content? Or is it about getting any non-error response from the target? Thoughts?

OJ:
Oh, and another thing. The intent with the `-v` flag, when doing verbose output, is to show messages when things fail, whereas this flag doesn't do anything at the moment. We should implement something that returns a missing result so that it can be logged with `-v`.

FireFart:
Currently it's a content based detection only. So if the new vhost returns something else than the default VHOST or the non exitent it will be reported because there is most likely some content to inspect. There is also the followredirect option if you want to follow 30x responses.
I will have a look into the `-v` and if there is any reasonable we can log 👍

FireFart:
@OJ implemented some verbose output, let me know what you think about it

gobustervhost/gobustervhost.go
```diff
@@ -109,7 +121,13 @@ func (v *GobusterVhost) Run(word string) ([]libgobuster.Result, error) {
...
 func (v *GobusterVhost) ResultToString(r *libgobuster.Result) (*string, error) {
 	buf := &bytes.Buffer{}
-	if _, err := fmt.Fprintf(buf, "Found: %s\n", r.Entity); err != nil {
+
+	statusText := "Found"
+	if v.globalopts.Verbose && r.Extra == "missed" {
+		statusText = "Missed"
+	}
+
+	if _, err := fmt.Fprintf(buf, "%s: %s (Status: %d) [Size: %d]\n", statusText, r.Entity, r.Status, *r.Size); err != nil {
```
OJ: I think it'd be better if:

* `Extra` is left alone in the result.
* The result is always passed through when Verbose is set.
* Any result is passed through with the `Status` set.
* The handler looks at the status to see if it's good or bad.

Verbose mode should also indicate "same as base" or something if the vhost responds with the same content.

Thoughts?

FireFart: The `Status` field is the return code which does not indicate success or failure (at least in this mode). I was thinking about adding a new field to the `Result` object indicating the result is a verbose one or smth like that.
What do you mean by `is passed through`?

FireFart: @OJ ok now I got what you meant. Implemented it for VHOSTS for now. Once this is merged we can update the other modes too to use the new (better) way

FireFart:
Merging this into v3 for public testing

Issue #61, opened 2017-10-08
Add dependency management (dep)


0xdevalias:
It would be nice to have the (soon to be official) supported package manager integrated.

* https://github.com/golang/dep

Definitely makes things easier when working on new features, and gives better support around managing dependency versions/branches/etc as required.

```
dep init
dep ensure
```

FireFart:
We are currently using gomodules

PR #111, opened 2018-10-21
Added Dependency Management (Issue #61)

kelsier27:
As referenced in Issue #61, I added dependency management using the `dep` package.


FireFart:
Hi @kelsier27! I added support for go modules in here https://github.com/OJ/gobuster/pull/117. I think this is more future proof than using the dep package

OJ:
Thanks @kelsier27 for this PR. As @FireFart has already mentioned, we've got another solution in place in our v3.0 working branch. So we'll respectfully close this PR. Thank you for your contribution though! It's great appreciated.

PR #117, opened 2018-12-04
Add support for go modules

firefart:
This adds support for go modules: https://github.com/golang/go/wiki/Modules

This makes it possible to build and use the repository outside of GOPATH (requires at least Go 1.11). As this is no runtime dependency the change should not affect other users.

The next release should be `v3.x.x` according to semantic versioning (we also need a tag named this way). See here for details:
https://github.com/golang/go/wiki/Modules#how-to-prepare-for-a-release

In the future we need to rename the import paths too if we release a major version

As always here is the link to the diff because there are a few branches to merge before this https://github.com/OJ/gobuster/compare/vhost...gomod

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ go:
   - "1.x"
   - "1.8"
   - "1.10.x"
+  - "1.11.x"
   - master
 
 script: make test
--- a/Makefile
+++ b/Makefile
@@ -8,11 +8,16 @@ current:
 	@go build -o ./gobuster; \
 	echo "Done."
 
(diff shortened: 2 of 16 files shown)


FireFart:
@OJ awesome thanks! It looks like I forgot a `v3` in an import statement (or it got lost during the merge) which I readded on the v3 working branch

OJ:
Cool thanks mate.

jojpyrat:
bonjour escusée moi de vous deranger jaurait besoin daide pour ce logitiel

Issue #75, opened 2018-02-26
Add JSON Output Feature


Ice3man543:
Hi,
First of all congratulations for building such an awesome tool. I mean it's damn fast and does the work nicely. 
I though it would be nice for gobuster to have a JSON output flag which would allow us to output the results in JSON format. This would allow for integration of gobuster in automated scripts. 
The current output format is not well enough for integration in automated scripts.

Thanks

mzpqnxow:
@Ice3man543 FYI I just started on this an hour ago.. right now it is only stdout for dir mode, but it shouldn't take more than another hour or two to have stdout+file output for dns and dir mode

I don't have the most time right now so if you want to work on it, check out https://github.com/mzpqnxow/gobuster/tree/json-logging

mzpqnxow:
Output currently looks like this:

```
$ ./gobuster.git -J -w ../raft-medium-words-lowercase.txt -u http://localhost -X OPTIONS -e 
{"Expanded":"true","Mode":"dir","Output file":"","Status":"starting","Status codes":"200,204,301,302,307","Threads":10,"URL":"http://localhost/","level":"info","msg":"Scan initializing","time":"2018-03-04T02:16:40-05:00"}
{"URL":"http://localhost/javascript","level":"info","msg":"HIT","size":-1,"status":301,"time":"2018-03-04T02:16:40-05:00"}
{"URL":"http://localhost/out","level":"info","msg":"HIT","size":-1,"status":200,"time":"2018-03-04T02:16:40-05:00"}
{"URL":"http://localhost/.","level":"info","msg":"HIT","size":-1,"status":200,"time":"2018-03-04T02:16:40-05:00"}
{"URL":"http://localhost/httpd","level":"info","msg":"HIT","size":-1,"status":200,"time":"2018-03-04T02:16:40-05:00"}
{"URL":"http://localhost/test2","level":"info","msg":"HIT","size":-1,"status":200,"time":"2018-03-04T02:16:40-05:00"}
{"URL":"http://localhost/test1","level":"info","msg":"HIT","size":-1,"status":200,"time":"2018-03-04T02:16:40-05:00"}
{"URL":"http://localhost/redirector","level":"info","msg":"HIT","size":-1,"status":200,"time":"2018-03-04T02:16:40-05:00"}
...
```

Ice3man543:
Hi @mzpqnxow, I will take a look at it.

Ice3man543:
[...]

PR #118, opened 2018-12-10
Add support for filtering results with a unique size for each status code

itsbriany:
When attempting to brute force web servers, they may return similar responses for different URLs. 

For example, a web server may redirect us to its homepage, giving us the same final result each time for multiple URLs. If we pipeline gobuster's output into other tools, there may be cases where we only care about URLs that point us to unique content.

For example, brute-forcing a web server with the following command:

`./gobuster -m dir -q -k -v -l -w wordlist.txt -u https://example.com`

And wordlist:

```
.
./lib
./bar
./car
../lib
../foo
../../bar
```

Will yield the following:

```
Missed: /../../bar (Status: 400) [Size: 357]
Missed: /../foo (Status: 400) [Size: 357]
Missed: /./car (Status: 404) [Size: 332]
Found: /./lib (Status: 301) [Size: 370]
Missed: /./bar (Status: 404) [Size: 332]
Missed: /../lib (Status: 400) [Size: 357]
Found: /. (Status: 200) [Size: 132]
```

Notice that the paths **../../bar** and **../foo** (likely) result in the same response since they have content of the same size.

There may be cases where I would like to have the following output instead:

```
Found: /./lib (Status: 301) [Size: 370]
Missed: /../foo (Status: 400) [Size: 357]
Missed: /./car (Status: 404) [Size: 332]
Found: /. (Status: 200) [Size: 132]
```

Since I only care about URLs that will give me a unique response.

I have implemented a patch that allows gobuster to filter results with a unique size for each status code to addresses the issue above.

If there is anything I have missed, I will be happy to address it in subsequent commits.

### DIFF
--- a/gobusterdir/gobusterdir.go
+++ b/gobusterdir/gobusterdir.go
@@ -117,7 +117,7 @@ func (d GobusterDir) ResultToString(g *libgobuster.Gobuster, r *libgobuster.Resu
 			}
 		}
 
-		if r.Size != nil {
+		if g.Opts.IncludeLength && r.Size != nil {
 			if _, err := fmt.Fprintf(buf, " [Size: %d]", *r.Size); err != nil {
 				return nil, err
 			}
--- a/libgobuster/http.go
+++ b/libgobuster/http.go
@@ -60,7 +60,7 @@ func newHTTPClient(c context.Context, opt *Options) (*httpClient, error) {
 	client.context = c
 	client.username = opt.Username
 	client.password = opt.Password
-	client.includeLength = opt.IncludeLength
+	client.includeLength = opt.IncludeLength || opt.UniqueResponseLength
 	client.userAgent = opt.UserAgent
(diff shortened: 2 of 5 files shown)


itsbriany:
I thought about this again and I think it actually might be better to keep the existing gobuster functionality, but instead write the output to JSON or XML and let another tool parse that output. I am investigating #75 to see if I can get gobuster to output JSON instead. 

If any of you think this is still a feature worth adding, let me know since my current solution in its current state will not function properly due to race conditions between threads. Additionally, synchronization checks are missing in my first commit. I also think that with my current solution, the amount of mutex locking/unlocking to achieve synchronization will also impact performance and hence should be designed differently.

IvoPereira:
@itsbriany I find a lot of use for this feature and I wouldn't see a reason why it wouldn't benefit the whole package itself instead of having a separate tool for it.

PR #121, opened 2018-12-18
Remove leading / from words

FireFart:
This PR removes leading `/` from words before processing them in dir mode. Some Wordlists (especially https://github.com/danielmiessler/SecLists) have leading slashes in them which results in URLs with double slashes

This PR depends on the other ones, here is this change only: https://github.com/OJ/gobuster/compare/gomod...slash


OJ:
Nope, and nope :) This tool isn't designed to munge inputs. It should take the input as is and not try to magically fix it. I'm really not keen on getting into a situation where we constantly try to adjust inputs.

People need to feed in the inputs they want and we treat it as is.

OJ:
Yeah that's exactly my thought. There are so many good text munging tools that can be used to parse and prepare the wordlist, I think we should leave it up to those. Thanks :)

Issue #115, opened 2018-11-07
Feature Request: Color results by respone code


eranchetz:
For example: 
2xx - green 
3xx - yellow
4xx - red 

we can use [aurora](https://github.com/logrusorgru/aurora) Sprintf [ResultToSring].(https://github.com/OJ/gobuster/blob/5ca1584c160e417bc7f7364f00bf3e756b0b2f25/gobusterdir/gobusterdir.go#L84)
@OJ wdyt? is that the proper place to implement it?

jaikishantulswani:
> For example:
> 2xx - green
> 3xx - yellow
> 4xx - red
> 
> we can use [aurora](https://github.com/logrusorgru/aurora) Sprintf [ResultToSring].(
> 
> [gobuster/gobusterdir/gobusterdir.go](https://github.com/OJ/gobuster/blob/5ca1584c160e417bc7f7364f00bf3e756b0b2f25/gobusterdir/gobusterdir.go#L84)
> 
> Line 84 in [5ca1584](/OJ/gobuster/commit/5ca1584c160e417bc7f7364f00bf3e756b0b2f25)
> 
>  func (d GobusterDir) ResultToString(g *libgobuster.Gobuster, r *libgobuster.Result) (*string, error) { 
> )
> @OJ wdyt? is that the proper place to implement it?

@eranchetz Is that working ?

eranchetz:
@jaikishantulswani I was hoping for a response before implementing, I will test it later this week and update.

eranchetz:
@jaikishantulswani https://github.com/OJ/gobuster/pull/122

jaikishantulswani:
Thank you @eranchetz  :)

OJ:
Thanks folks. Yes there's a PR open for it, but needs a little bit more work before it can be landed. Cheers!

firefart:
This is implemented on the `dev` branch for now

PR #122, opened 2019-01-14
Add colors to gobusterdir (#115)

eranchetz:
This relates to #115.
If there is a more efficient way to determine if a request is 200, I would be more than happy to make the changes.

### DIFF
--- a/gobusterdir/gobusterdir.go
+++ b/gobusterdir/gobusterdir.go
@@ -7,6 +7,8 @@ import (
 
 	"github.com/OJ/gobuster/libgobuster"
 	"github.com/google/uuid"
+
+	. "github.com/logrusorgru/aurora"
 )
 
 // GobusterDir is the main type to implement the interface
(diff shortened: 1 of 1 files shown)


gobusterdir/gobusterdir.go
```diff
@@ -87,11 +89,11 @@ func (d GobusterDir) ResultToString(g *libgobuster.Gobuster, r *libgobuster.Resu
 	// Prefix if we're in verbose mode
 	if g.Opts.Verbose {
 		if g.Opts.StatusCodesParsed.Contains(r.Status) {
-			if _, err := fmt.Fprintf(buf, "Found: "); err != nil {
+			if _, err := fmt.Fprintf(buf, Sprintf(Green("Found: "))); err != nil {
 				return nil, err
 			}
 		} else {
-			if _, err := fmt.Fprintf(buf, "Missed: "); err != nil {
+			if _, err := fmt.Fprintf(buf, Sprintf(Red("Missed: "))); err != nil {
```
OJ: I think I'd rather see `Missed` outputted in orange or yellow. It's not exactly an error.

gobusterdir/gobusterdir.go
```diff
@@ -112,7 +114,11 @@ func (d GobusterDir) ResultToString(g *libgobuster.Gobuster, r *libgobuster.Resu
 		}
 
 		if !g.Opts.NoStatus {
-			if _, err := fmt.Fprintf(buf, " (Status: %d)", r.Status); err != nil {
+			statusWithColor := Sprintf(Brown(r.Status))
+			if r.Status == 200 {
+				statusWithColor = Sprintf(Green(r.Status))
```
OJ: I think it's unnecessary to have different colours specifically for 200. If you get a hit, based on the requested status codes, that's success and should be green.

OJ:
Also, can you please not submit pull requests from your master branch in future? Best to do it from a feature branch.

If you could rebase this against the v3.0-working branch also that'd be super helpful. Thanks!

Anon-Exploiter:
@OJ @eranchetz any progress on the **colored output** of gobuster? :)

eranchetz:
@Anon-Exploiter this have become more complicated than I originally thought.
I think I will forgo this :(

Issue #53, opened 2017-08-23
Minimize content downloaded for each checked URL


0xdevalias:
I haven't confirmed this, but the feel I got from looking at the code is that currently gobuster downloads the entire page for each checked URL.

If that is the case, there are a few methods I was thinking of that may help to speed this up:
* Using the range header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests
* Supporting HEAD/OPTIONS requests (https://github.com/OJ/gobuster/pull/44)

On the topic of using the range header, it could be that some tests are done at the beginning to see if it is supported (similar to the 'wildcard response found' check) to ensure the server supports ranges.

If there is support for it, gobuster could send the range header to limit the request to a reasonably small size (maybe customizable from flags with a sensible default)

If it's not supported by the server, maybe gobuster could forcibly close the connection after reading X bytes as a fallback.

If there isn't already a timeout set for a connection that 'never returns', that could be implemented in a similar way, again customisable.

There may be other optimisations that could be done in a similar vain to this, but for now, these are the ones coming to mind.

mzpqnxow:
@0xdevalias I would caution against using HEAD as there are many common server configurations that either explictly reject HEAD, reject it silently with a 404, or behave in some otherwise misleading way, leading mainly to false negatives.

Just my $.02

0xdevalias:
Yeah, that is definitely a good point. I’d sort of half considered it, but probably didn’t really articulate that.

FireFart:
You can now specify a custom http verb in the `v3.1-cleaned` branch so I classify this as done :)

PR #123, opened 2019-02-04
Add option to use HEAD requests

int0x80:
The intent of this patch is to improve speed by utilizing the HEAD method of HTTP. The current default/only option for gobuster is to make GET requests to the target service. This incurs a fair amount of overhead as gobuster will receive the full contents of the URI in addition to the HTTP response headers. By using HEAD, only the response headers will be returned, from which the HTTP code and Content-Length can be parsed. In the event that the HEAD method is not supported by the target, a `405 Method Not Allowed` or `501 Not Implemented` response will disable the use of HEAD requests and revert to the original behavior of GET requests.

Notably @0xdevalias requests this feature and @mzpqnxow warns against this approach in [#53](https://github.com/OJ/gobuster/issues/53). Accordingly we only opt for HEAD requests when specified by the user via the `-H` or `--head` command line option added with this diff.

### DIFF
--- a/cli/cmd/dir.go
+++ b/cli/cmd/dir.go
@@ -169,6 +169,11 @@ func parseDirOptions() (*libgobuster.Options, *gobusterdir.OptionsDir, error) {
 		return nil, nil, fmt.Errorf("invalid value for addslash: %v", err)
 	}
 
+	plugin.HeadRequests, err = cmdDir.Flags().GetBool("head")
+	if err != nil {
+		return nil, nil, fmt.Errorf("invalid value for head: %v", err)
+	}
+
 	// Prompt for PW if not provided
 	if plugin.Username != "" && plugin.Password == "" {
 		fmt.Printf("[?] Auth Password: ")
@@ -212,6 +217,7 @@ func init() {
 	cmdDir.Flags().BoolP("insecuressl", "k", false, "Skip SSL certificate verification")
 	cmdDir.Flags().BoolP("addslash", "f", false, "Apped / to each request")
 	cmdDir.Flags().BoolP("wildcard", "", false, "Force continued operation when wildcard found")
+	cmdDir.Flags().BoolP("head", "H", false, "Use HEAD requests if possible")
 
 	if err := cmdDir.MarkFlagRequired("url"); err != nil {
 		log.Fatalf("error on marking flag as required: %v", err)
--- a/gobusterdir/http.go
+++ b/gobusterdir/http.go
@@ -69,7 +69,12 @@ func newHTTPClient(c context.Context, opt *OptionsDir) (*httpClient, error) {
 
 // MakeRequest makes a request to the specified url
 func (client *httpClient) makeRequest(fullURL, cookie string) (*int, *int64, error) {
-	req, err := http.NewRequest(http.MethodGet, fullURL, nil)
+	// attempt a HEAD request if specified by the user 
+	if opt.HeadRequests {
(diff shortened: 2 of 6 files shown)


gobusterdir/http.go
```diff
@@ -68,9 +68,14 @@ func newHTTPClient(c context.Context, opt *OptionsDir) (*httpClient, error) {
 }
 
 // MakeRequest makes a request to the specified url
-func (client *httpClient) makeRequest(fullURL, cookie string) (*int, *int64, error) {
+func (client *httpClient) makeRequest(fullURL, cookie string, opt *OptionsDir) (*int, *int64, error) {
```
0xdevalias: If you’re passing in the full options now, I wonder if it makes sense to pull the cookie out of this function params, and just extract it from the options inside this function as needed

gobusterdir/http.go
```diff
@@ -103,6 +108,11 @@ func (client *httpClient) makeRequest(fullURL, cookie string) (*int, *int64, err
 		return nil, nil, err
 	}
 
+	// disable HEAD requests if the target does not support the method
+	if opt.HeadRequests && (resp.StatusCode == 405 || resp.StatusCode == 501) {
```
0xdevalias: Is it worth noting this to the user somehow via logging/similar?

cli/cmd/dir.go
```diff
@@ -212,6 +217,7 @@ func init() {
 	cmdDir.Flags().BoolP("insecuressl", "k", false, "Skip SSL certificate verification")
 	cmdDir.Flags().BoolP("addslash", "f", false, "Apped / to each request")
 	cmdDir.Flags().BoolP("wildcard", "", false, "Force continued operation when wildcard found")
+	cmdDir.Flags().BoolP("head", "H", false, "Use HEAD requests if possible")
```
OJ: Given the general use case, it might be worth defaulting to `HEAD` over `GET` and falling back if it doesn't work (or if the user asks for the length of the response). But given that we have the need/desire to handle other verbs, we might need to generalise this.

gobusterdir/http.go
OJ: +1 to this. Pass in all options instead.

gobusterdir/http.go
OJ: Agreed here too. Definitely log something that indicates the fallback.

gobusterdir/options.go
```diff
@@ -27,6 +27,7 @@ type OptionsDir struct {
 	UseSlash          bool
 	IsWildcard        bool
 	WildcardForced    bool
+	HeadRequests      bool
```
OJ: So before we go ahead with this, should we discuss the handling of a wider range of verbs?

gobusterdir/http.go
int0x80: Makes sense, I think this adds to improve the user experience.

gobusterdir/http.go
int0x80: Indeed it seemed redundant to leave the cookie parameter when it was already being passed in along with the options. I was going for a light touch on the codebase, hence the initial decision to simply add one parameter.

cli/cmd/dir.go
int0x80: :+1: for utilizing other verbs. It appears `-m` is available in `dir` mode, perhaps usage could look something like:

```
-m, --method               HTTP method/verb to use for request (GET, HEAD, etc.)
```

Should we also consider a flag that persists using the specified HTTP verb regardless of the web server response?

PR #130, opened 2019-03-06
Unnecessary conversion.

d-kuro:


FireFart:
This is needed if you compile it on windows (sadly). We removed it once and then compilation on other OS failed

d-kuro:
I didn't know, thanks!

PR #132, opened 2019-03-22
Bust using placeholder strings.

leonjza:
# proposal

Reimplementation of #71.

Add the ability to use a placeholder value to replace parts of a URL using a wordlist. The same placeholder string may be used in multiple locations, ultimately allowing for complex busting.

Example replacements:

```text
http://127.0.0.1/zFOOa -> http://127.0.0.1/zword1a
http://127.0.0.1/zFOOa -> http://127.0.0.1/zword2a
http://127.0.0.1/zFOOa/FOO -> http://127.0.0.1/zword2a/word2
http://127.0.0.1/FOO.txt -> http://127.0.0.1/word3.txt
```

# sample usage

```bash
$ go run main.go -q dir --url http://127.0.0.1:58863/aFOOzFOO -w common.txt -t 1 -S FOO
/netcat (string replacement requested as: http://127.0.0.1:58863/anetcatznetcat) (Status: 301)
/zip (string replacement requested as: http://127.0.0.1:58863/azipzzip) (Status: 200)
```

Regardless if the problems mentioned below, I believe this to be a useful feature nonetheless.

# problems

1. A trailing slash gets appended in `GobusterDir.PreRun()` and causes misses on some web servers where the target URI is a file (not a dir). I know this is added in the `dir` subcommand, but I added a strip for this slash and only add the slash when `-f` is specified again. However, this causes an inverse effect when the original URL actually has a `/` and `-f` is not specified. I suspect this may need some different handling overall. I am also happy to remove that and keep the intended `dir` context of the subcommand.

2. Output by default is not very useful given the nature when a replace method is used. I added a more verbose blurp as I am not sure if editing the original URL is really a good idea. That being said, `-v` and `-e` seem equally odd when it comes to output because of this.

### DIFF
--- a/cli/cmd/dir.go
+++ b/cli/cmd/dir.go
@@ -129,6 +129,15 @@ func parseDirOptions() (*libgobuster.Options, *gobusterdir.OptionsDir, error) {
 		return nil, nil, fmt.Errorf("invalid value for proxy: %v", err)
 	}
 
+	plugin.StringReplace, err = cmdDir.Flags().GetString("stringreplace")
+	if err != nil {
+		return nil, nil, fmt.Errorf("invalid value for stringreplace: %v", err)
+	}
+
+	if plugin.StringReplace != "" && !strings.Contains(plugin.URL, plugin.StringReplace) {
+		return nil, nil, fmt.Errorf("stringreplace token '%s' not found in url", plugin.StringReplace)
+	}
+
 	plugin.Timeout, err = cmdDir.Flags().GetDuration("timeout")
 	if err != nil {
 		return nil, nil, fmt.Errorf("invalid value for timeout: %v", err)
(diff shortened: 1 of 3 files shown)


OJ:
Thanks @leonjza I'll get stuck in shortly.

leonjza:
Not worth it?

OJ:
Umm... I did not manually close this. Something's up.

OJ:
Why does it say unknown repo? Did you ditch your fork?

leonjza:
Yeah I clicked the shiny "delete fork" button considering this was closed.

OJ:
I hope that wasn't a ragequit. I promise I didn't do that manually.

leonjza:
Nah. Simply figured a closed PR meant it wont land, and as a result I saw no need to keep my fork. I can PR it again sometime if needed.

PR #135, opened 2019-04-02
read urls from file

Margular:
add support for read urls from file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ _testmain.go
 
 gobuster
 build
+.idea/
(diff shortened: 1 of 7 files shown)


FireFart:
Hey @Margular,
thanks for your PR! Our current development branch is `v3.0-working` so new PR should target this branch instead of `master`.

I have a few comments:
- what's the advantage of looping over the file with bash and launch multiple gobusters?
- please don't introduce new dependencies `go-multierror`
- please don't change the import paths, they should still be `github.com/OJ/gobuster/...`
- please create a feature branch for merge requests on your side, currently you are using the `master` branch
- @OJ what do you think of this?

Margular:
Yeah, but I think group scan is't possible when using bash, I mean multi hosts parallel scanning

PR #146, opened 2019-05-24
V3.0 working - Updates

mattburch:
Some of the default I/O STDERR messages made it difficult to thread multiple background executions of the tool. With the I/O adjustments, it is very easy to monitor the state of multiple requests.

```
$ ./gobuster dir -w /opt/lists/directory-list-lowercase-2.3-medium.txt -e -z -t 5 -k -u http://10.10.10.10/ 2>/dev/null
Gobuster v3.0.0 (beta)
by OJ Reeves (@TheColonial) & Christian Mehlmauer (@_FireFart_)
[+] Url:            http://10.10.10.10/
[+] Threads:        5
[+] Wordlist:       /opt/lists/directory-list-lowercase-2.3-medium.txt
[+] Status codes:   200,204,301,302,307,401,403
[+] User Agent:     gobuster/3.0.0 (beta)
[+] Expanded:       true
[+] Timeout:        10s
2019-05-24 09:30:42 Starting gobuster
[*] Wildcard Detected: http://10.10.10.10/ => 302
```

### DIFF
--- a/cli/cmd/dir.go
+++ b/cli/cmd/dir.go
@@ -25,7 +25,9 @@ func runDir(cmd *cobra.Command, args []string) error {
 	}
 
 	if err := cli.Gobuster(mainContext, globalopts, plugin); err != nil {
-		if err == gobusterdir.ErrWildcard {
+		if plugin.WildCard.ErrorCheck(err) {
+			url, code := plugin.WildCard.GetDir()
+			fmt.Printf("[*] Wildcard Detected: %s => %v\n", url, code)
 			return fmt.Errorf("the server returns the same status code for every request. To force processing of Wildcard responses, specify the '--wildcard' switch")
 		}
 		return fmt.Errorf("error on running goubster: %v", err)
--- a/cli/cmd/dns.go
+++ b/cli/cmd/dns.go
@@ -26,7 +26,9 @@ func runDNS(cmd *cobra.Command, args []string) error {
 	}
 
 	if err := cli.Gobuster(mainContext, globalopts, plugin); err != nil {
-		if err == gobusterdns.ErrWildcard {
+		if plugin.WildCard.ErrorCheck(err) {
+			domain, ip := plugin.WildCard.GetDNS()
+			fmt.Printf("[*] Wildcard Detected: %s => %v\n", domain, ip)
 			return fmt.Errorf("the DNS Server returned to same IP for every domain. To force processing of Wildcard DNS, specify the '--wildcard' switch")
 		}
 		return fmt.Errorf("error on running goubster: %v", err)
--- a/cli/gobuster.go
+++ b/cli/gobuster.go
@@ -9,7 +9,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/OJ/gobuster/v3/helper"
 	"github.com/OJ/gobuster/v3/libgobuster"
 )
 
@@ -123,7 +122,7 @@ func Gobuster(prevCtx context.Context, opts *libgobuster.Options, plugin libgobu
 		}
 		fmt.Println(c)
 		ruler()
-		helper.MsgTimeStamp("Starting gobuster")
+		libgobuster.MsgTimeStamp("Starting gobuster")
 		ruler()
 	}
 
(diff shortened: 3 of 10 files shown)


FireFart:
Hi,
what's wrong with log.Println? Both should print to stdout?

mattburch:
If multiple background processes are executed, having time stamps is
valuable to obtain the information. Println is an option as well but you
loose variable injection in the content.

mattburch:
Typically I would execute the tool within a screen session and have
multiple threads executed against various hosts. Having stdout responses,
that are timestamped, allows me to review the session log and identify
start/stops and the status responses.

FireFart:
I don't get the point sorry, log.Println already outputs the current timestamp on my machine?

mattburch:
The tool also can produce a lot of error stderr response content due to
standard net.Dial requests. The wildcard indicator is important, similar to
start/stop notifications and can get lost in 2>/dev/null for background
process execution. I agree stderr should be used for error conditions, but
WildCard flags are an app defined construct not a typical error condition.
Splitting the content into a stdout status condition along with an stderr
response allows for the conditional state of tool execution to be monitored
when error content is sent to /dev/null.

FireFart:
The wildcard responses will be implemented again, but you also changed the log.Println to a fmt.Println with a timestamp which should not be necessary as it's already there

mattburch:
Start/Stop notifications were sent to stderr via a log statement. I
modified these to fmt.Print so they would be issued against stdout, but
that lost the timestamp which was a history tracker of when the tool
started and stopped. This tool is leveraged for pentesting efforts and some
clients require and indication when activities start/stop; which is not
possible to produce without a timestamp. Hence the stdout function of
providing a timestamp to stdout message content.

The Wildcard changes were submitted with a different git init comment.

mattburch:
Based on the original log.Print statement, it appeared applying a timestamp
was intended functionality. FMT does not produce the same interface logging
method, hence the function creation within libgobuster help file.

FireFart:
So it's all about stderr I see. The fix for this would be another one, I will take a look at it thanks

FireFart:
@mattburch I added the features to the v3 branch. Would you mind giving it a test and if this works out for you?

mattburch:
So couple of notes. When the tool finishes:
```
2019/05/24 14:44:07 Finished
```

I modified this response to look like
```
2019/05/24 14:44:07 Finished - <URL|Domain>
```

This output provides more relevance to what task had actually completed if
this is executed as a background job.

Also the WildCard error output, below, still follows the stderr I/O channel
and can get lost if filtering error content.
```
Error: the server returns the same status code for every request.
http://10.10.10.10/0f4a110b-9953-402e-85e3-27dcb6f792ff => 302. To force
processing of Wildcard responses, specify the '--wildcard' switch
```

FireFart:
Personally I would redirect each domain in a seperate file to seperate the domains. I think putting it in the finished message will distract normal users only scanning one domain / url.

For the wildcard message: This is an error and the program exits afterwards so this will always be in stderr, sorry

mattburch:
I understand the suggestion about piping the contents into a file, however
it appears this does not actually resolve the issue. Output logging of the
command execution does not capture start/stop times or input parameters.
This makes it difficult to present an accurate run book of activities
performed during an assessment exercise. This also leads back into the
previously mentioned item, that if your command is used to thread multiple
tasks, there isn't functionality to association the completion of one task
over the other.

In regards to the wildcard discussion, I understand the notification state
which initiates a programmatic closure of the application to be considered
an error response (stderr), however I don't see why you would be opposed to
following the metasploit status notification indicators [+] [-] [*] that a
possible wildcard condition was detected "[*] WildCard response identified".

FireFart:
We can't fix everything in the tool, it should be kept simple and you can build wrappers around it to call it automatically. We also try not to do any wordlist mangling and stuff like this because it should be done outside of the tool. With v3 you can even use and import libgobuster in your own code and make use of it's functionality and call it many times in a loop.
The start and stop times are already printed out so there is a clear indication when the command started and when it finished. For the input parameters you could also log them in the wrapper.
The `--wildcard` is also clearly an error as the program exits when this happens and does not continue. If you provide the wildcard switch to the programm this message will never show up.

PR #152, opened 2019-06-22
Added an option to add a delay to threads

0x6368:


cli/cmd/root.go
```diff
@@ -109,6 +119,7 @@ func configureGlobalOptions() {
 }
 
 func init() {
+	rootCmd.PersistentFlags().IntP("delay", "d", 0, "Number of milliseconds each thread waits between requests")
```
FireFart: this can be a `DurationP` so you don't need to convert it

libgobuster/libgobuster.go
```diff
@@ -113,6 +114,8 @@ func (g *Gobuster) worker(wordChan <-chan string, wg *sync.WaitGroup) {
 					g.resultChan <- r
 				}
 			}
+
+			time.Sleep(time.Duration(g.Opts.Delay) * time.Millisecond)
```
FireFart: this should be smth like this so ctrl+c works correctly:
```golang
select {
  case <-g.context.Done():
    return
  case <-time.After(g.Opts.Delay):
    return
}
```

libgobuster/options.go
```diff
@@ -10,6 +10,7 @@ type Options struct {
 	Quiet          bool
 	WildcardForced bool
 	Verbose        bool
+	Delay          int
```
FireFart: This should be a `time.Duration`

gobusterdir/gobusterdir.go
```diff
@@ -242,10 +242,15 @@ func (d *GobusterDir) GetConfigString() (string, error) {
 	if _, err := fmt.Fprintf(tw, "[+] Url:\t%s\n", o.URL); err != nil {
 		return "", err
 	}
+
 	if _, err := fmt.Fprintf(tw, "[+] Threads:\t%d\n", d.globalopts.Threads); err != nil {
 		return "", err
 	}
 
+	if _, err := fmt.Fprintf(tw, "[+] Delay:\t%dms\n", d.globalopts.Delay); err != nil {
```
FireFart: we should only print this out if it is set

gobusterdns/gobusterdns.go
```diff
@@ -173,6 +173,10 @@ func (d *GobusterDNS) GetConfigString() (string, error) {
 		return "", err
 	}
 
+	if _, err := fmt.Fprintf(tw, "[+] Delay:\t%d\n", d.globalopts.Delay); err != nil {
```
FireFart: we should only print this out if it is set

gobustervhost/gobustervhost.go
```diff
@@ -143,10 +143,15 @@ func (v *GobusterVhost) GetConfigString() (string, error) {
 	if _, err := fmt.Fprintf(tw, "[+] Url:\t%s\n", o.URL); err != nil {
 		return "", err
 	}
+
 	if _, err := fmt.Fprintf(tw, "[+] Threads:\t%d\n", v.globalopts.Threads); err != nil {
 		return "", err
 	}
 
+	if _, err := fmt.Fprintf(tw, "[+] Delay:\t%d\n", v.globalopts.Delay); err != nil {
```
FireFart: we should only print this out if it is set

FireFart:
please also create a feature branch for your pull request instead of creating it from `master`

0x6368:
Thanks for the feedback!
I changed the delay type to time.Duration and created a new pull request from a feature branch: https://github.com/OJ/gobuster/pull/153

Issue #147, opened 2019-05-29
Delay option?


Athrekas:
Hello, I'd like to know if it's possible to set a delay per request, or at least knowing what the default delay is.

For example on dirsearch I can do "-t 3 -s 0.350" (3 threads, 350 miliseconds per request).

Thank you very much.

FireFart:
Hi,
no currently not but we could implement it (at least per thread)

Athrekas:
Awesome! Thank you for the quick response!

FireFart:
Sorry this was closed by accident by merging. The reverted implementation does not really work so we need to implement a sleep per thread here

FireFart:
this will be released in the next version as a `--delay` parameter

PR #153, opened 2019-06-23
Added an option to add a delay to threads

0x6368:
Cross reference to #147 as this was requested there


FireFart:
Thanks this looks good from here!

PR #163, opened 2019-07-14
Fixed gobuster spelling error

nbulischeck:
After running into an error, I noticed that "Gobuster" was misspelled. After running a quick recursive grep I found that this mistake was repeated quite a few times. This is a small patch to fix those misspellings.


FireFart:
nice catch thanks!

PR #173, opened 2019-08-15
v3.1

FireFart:
## Main features
- Added S3 bucket enumeration. Currently we only check for existence and if the bucket is listable. To automatically pre and post-pend stuff like -dev and -prod take a look at the new permutations feature
- custom HTTP methods are now supported
- exclude-length parameter to exclude specific content-lengths
- added a fuzzer
- random useragents
- patterns (See Readme.md)
Example pattern file:
```text
{GOBUSTER}Partial
{GOBUSTER}Service
PRE{GOBUSTER}POST
{GOBUSTER}-prod
{GOBUSTER}-dev
```

## Internal Stuff
- Reduced types inside gobuster by reusing them
- makes the helper function names more generic
- extracted the bare basic http options out so we can work with a subset of them also
- cleans up the HTTP package a lot and remove duplicated code
- get rid of ansi control characters and fix messed up output in some cases


FireFart:
closed in favor of https://github.com/OJ/gobuster/pull/225

PR #177, opened 2019-09-12
dir wildcard and levenshtein distance

fopina:
The wildcard response detection is quite useful, no one needs a bunch of false positives to go through.
But as we detect wildcard we can also try to use that as baseline to compare with future responses, hence `--wildcard-detect` option.

Using [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance), we can compare each response (with a wildcard-possible status code) body and compare it to our baseline and then discard those results.

## Pros

* Instead of getting all the false positives with `--wildcard`, we only get true positives

## Cons

* Each request will always read the body (taking longer)
* If a website uses a big base template and a true positive and a false positive differ very little, true positives will be discarded

## Testing

Used the following dockerfile for a dummy server 

```
FROM nginx:alpine

ARG NOTFOUND_STATUS=200

RUN echo 'not found' > /usr/share/nginx/html/404.html
RUN echo 'gotme' > /usr/share/nginx/html/right
RUN sed -i "s/#error_page  404/error_page 404 =${NOTFOUND_STATUS}/" /etc/nginx/conf.d/default.conf
```

Built it with `docker build -t bustertest -f Dockerfile.nginx.test .`
Ran it with `docker run --rm -p8888:80 bustertest`
Busted it with `go run main.go dir -u http://localhost:8888/ -w test.txt --wildcard-detect`

## Notes

* `--wildcard-detect` was a random option name just to get it going, if the idea LGTY, I'll just rename as you find it better
* `_test.go` and documentation updates will follow if PR gets the 👍


OJ:
Thanks for the PR @fopina, I'm hoping to get to reviewing this very soon.

fopina:
Cool. I have to try Humming distance as this one slows the scan down too much but I haven’t found the time to test it yet... or maybe you have some suggestion on something to be improved :)

fopina:
Any review comments?

My highlights so far:
* does it make sense to be separate option than —wildcard or replace it?
* big wildcard responses increase scanning time immensely due to levenshtein being slow, consider another one such as humming?
* if keeping as separate option, better name for it?

PR #182, opened 2019-10-01
Added MTLS support

fdcarl:
Adding MTLS support for both the dir and vhost commands. 

- Adds the -C switch to specify the MTLS certificate file
- Adds the -K switch to specify the MTLS certificate key files


OJ:
Nice! This looks like a super handy addition. Thank you :)

FireFart:
What is this for? Are those client certificates? We should definitely specify this in the module options more clearly

firefart:
thanks this was added to the dev branch!

PR #187, opened 2019-10-21
Add support for Host header

Vigorwasp:
If I want to provide an IP and give a custom Host header (like in cases of hosts behind Cloudflare for example), I have to provide a custom Host header. However, it's passed into the headers argument and is ignored by Go's http lib. See: https://github.com/golang/go/issues/7682

This issue can usually be worked around by adding the IP to your hosts file, but when proxying the requests, this won't help. I have fixed this issue in a way that every Host header passed to `-H` will be passed as `.Host` in the requests.

Also, in `addCommonHTTPOptions`, the function call to mark the URL flag as required was written incorrectly. It was called on the `cmdDir` instance everytime, and therefore `gobuster vhost --wordlist file` would not detect the URL flag was missing and ran anyway (with an empty url)


FireFart:
@Vigorwasp thanks for your contribution. Our current development branch is `v3.1-working`. Can you please rebase your PR to this branch?

FireFart:
I think this might also need a rebase the the `cmdDir` error is aleady fixed on that branch

Vigorwasp:
Working on it.

Vigorwasp:
Should be sorted now. I kept the `Host` option in the `HTTPOptions` struct and and not in `BasicHTTPOptions`. The way `parseCommonHTTPOptions` is written atm makes it harder to parse the 'Host' header out of the -H argument without breaking the function's idea (BasicHTTPOptions is parsed in a different function - parseBasicHTTPOptions).

Seems to be working the same way, and gives precedence to the host argument given in the request function, so it doesn't break the vhost functionality.

`gobuster dir --url http://localhost -H 'Host: site.com'` should work as expected.

FireFart:
@Vigorwasp looks like there are still some errors
Screenshot 2019-10-21 at 19 41 09

Vigorwasp:
@FireFart Not sure why, no conflicts on my end right now.

FireFart:
Looks like because your branch is based on master, not on v3.1-working:
Screenshot 2019-10-21 at 19 47 20

Vigorwasp:
Not sure what's wrong - looks OK on my end.


Are you able to see what are the conflicts?

FireFart:
I would say it's because your branch is based of master and not the working branch but I can only see the grey button on github. Maybe it's best to create a new PR with the changes

Vigorwasp:
@FireFart I'll try to recreate, see if it works then.

PR #188, opened 2019-10-21
Add support for Host header

yanir3:
If I want to provide an IP and give a custom Host header (like in cases of hosts behind Cloudflare for example), I have to provide a custom Host header. However, it's passed into the headers argument and is ignored by Go's http lib. See: https://github.com/golang/go/issues/7682

This issue can usually be worked around by adding the IP to your hosts file, but when proxying the requests, this won't help. I have fixed this issue in a way that every Host header passed to `-H` will be passed as `.Host` in the requests.


FireFart:
You created the same branch as before. You can see here it's still based off of master:
https://github.com/Vigorwasp/gobuster/network
You need to create a new branch from v3.1-working and apply the patches to it

Vigorwasp:
@FireFart My branch is merged with v3.1-working (which is obviously based off master). I see no conflicts and neither does GitHub's compare tool - https://github.com/OJ/gobuster/compare/v3.1-working...Vigorwasp:bugfix/host-header-and-flags

libgobuster/http.go
```diff
@@ -135,6 +137,10 @@ func (client *HTTPClient) makeRequest(fullURL, host string, data io.Reader) (*ht
 		req.Header.Set("Cookie", client.cookies)
 	}
 
+	if client.host != "" {
```
FireFart: there is already a host parameter passed in. This should be used instead of introducing new struct fields

FireFart:
> @FireFart My branch is merged with v3.1-working (which is obviously based off master). I see no conflicts and neither does GitHub's compare tool - [v3.1-working...Vigorwasp:bugfix/host-header-and-flags](https://github.com/OJ/gobuster/compare/v3.1-working...Vigorwasp:bugfix/host-header-and-flags)

yeah but github means it can not be merged so there is something wrong

libgobuster/http.go
Vigorwasp: Should be good now

n-thumann:
Any chance you can fix the merge conflict @yanir3? ✌️ I would love to see this feature!

yanir3:
@n-thumann fixed, but it doesn't seem like it will be merged soon.

n-thumann:
> @n-thumann fixed, but it doesn't seem like it will be merged soon.

Thank´s for the fast fix 👏 Why is that?

yanir3:
> > @n-thumann fixed, but it doesn't seem like it will be merged soon.
> 
> Thank´s for the fast fix 👏 Why is that?

Last version release was almost a year ago.

n-thumann:
I´m running this branch for a while now and it works like a charm 👏 
Maybe @FireFart can have a look, now that the merge conflict is solved?

PR #200, opened 2019-12-31
Add Feature to discover backup files.

IppSec:
Added the -d flag to DiscoverContent.  This will take all files with a 200 HTTP Response and then mutate the file to find backup copies.

**Example:** 
index.php ->
* index.php~
* index.php.bak
* index.bak
* .index.php.swp

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,7 @@ Flags:
   -a, --useragent string              Set the User-Agent string (default "gobuster/3.0.1")
   -U, --username string               Username for Basic Auth
       --wildcard                      Force continued operation when wildcard found
+  -b  --backuphunt                    Upon 200 searches for backup files
 
 Global Flags:
   -z, --noprogress        Don't display progress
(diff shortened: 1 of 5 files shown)


FireFart:
@IppSec thanks for your contribution! I think we already have this implemented in the `v3.1-working` branch over here https://github.com/OJ/gobuster/tree/v3.1-working with the new `--pattern` option:
https://github.com/OJ/gobuster/tree/v3.1-working#patterns

Does that fit?

IppSec:
I think they are different but I don’t fully understand the pattern option.  

The main difference is this generates the wordlist based upon found files, whereas pattern applies it to everything in the dirlist.  Additionally, I don’t think patterns supports removing extension then appending items.

FireFart:
Oh ok so it only searches backups for found files, missed that. Can you please rebase the branch on the v3.1-working branch as this is the current development version? Thanks!

IppSec:
Created #201 to rebase against my v3.1 branch.

PR #201, opened 2020-01-03
Added Discovery Option to DIR Mode

IppSec:
NOTE: This is a duplicate from #200, I just moved and tested it over to the v3.1-working branch.

Added the -d flag to DiscoverContent. This will take all files with HTTP FOUND Status Codes and then mutate the file to find backup copies.

Example:
index.php ->

index.php~
index.php.bak
index.bak
.index.php.swp


FireFart:
Thanks for the contribution! Looks good so far

PR #141, opened 2019-04-18
Output JSON

kulinacs:
Adds a JSON output flag. Implemented in line with the file output functionality.

Fixes #75 and addresses changes requested in #119

PR #205, opened 2020-02-08
Support JSON output

kulinacs:
Supports file output to JSON for easy import into other tooling. Rework of #141

### DIFF
--- a/cli/gobuster.go
+++ b/cli/gobuster.go
@@ -22,7 +22,6 @@ func banner() {
 type outputType struct {
 	Mu              *sync.RWMutex
 	MaxCharsWritten int
-	JSON            bool
 }
 
 // right pad a string
@@ -52,8 +51,8 @@ func resultWorker(g *libgobuster.Gobuster, filename string, wg *sync.WaitGroup, 
 		defer f.Close()
 	}
 
-	if f != nil && output.JSON {
-		writeToFile(f, "[")
+	if f != nil && g.Opts.JSON {
+		writeToFile(f, "[", false)
 	}
 
 	first := true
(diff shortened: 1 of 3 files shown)


cli/gobuster.go
```diff
@@ -126,7 +148,7 @@ func progressWorker(c context.Context, g *libgobuster.Gobuster, wg *sync.WaitGro
 }
 
 func writeToFile(f *os.File, output string) error {
-	_, err := f.WriteString(fmt.Sprintf("%s\n", output))
+	_, err := f.WriteString(output)
```
FireFart: This kills normal text output to a file because you are removing the new line character. You could pass an option to `writeToFile` if we should include the newline character or not

cli/gobuster.go
```diff
@@ -177,6 +199,7 @@ func Gobuster(prevCtx context.Context, opts *libgobuster.Options, plugin libgobu
 	o := &outputType{
 		Mu:              outputMutex,
 		MaxCharsWritten: 0,
+		JSON:            opts.JSON,
```
FireFart: no need for that, you can access the property in the result worker from the gobuster variable.

cli/gobuster.go
kulinacs: Updated to include an optional flag to append a newline character.

cli/gobuster.go
kulinacs: removed

kulinacs:
@FireFart requested changes made

FireFart:
I'm not really happy with the output format. Currently we marshal the whole `Result` type struct into a json, but the to string method does a lot more based on the content and plugin. So maybe we should add a `ToJSON` method that returns formatted json data in the cirrect format (like the `ToString`)

Issue #185, opened 2019-10-13
Progress Request Count Visual Bug - Minor


ngkogkos:
Hello,

When gobuster shows progress for example `5 out of 500` it does not calculate the extra requests performed by added extensions.

So if I feed it with a wordlist of 10 files and give it 4 extra extension, it will still show:

```
1 out of 10
2 out of 10
...
```

instead of making the overall number of requests 50, so `1 out of 50`.

It's mainly a visual bug I think and doesn't matter much, but good for improvement.

FireFart:
this is now fixed in the `v3.1-cleaned` branch

PR #211, opened 2020-03-11
#185 progress counts all requests

layderv:
Issue: https://github.com/OJ/gobuster/issues/185

I tried to modify as few files as possible. Let me know if I should consider another approach.


FireFart:
Hi @layderv ,
thanks for the PR, our dev branch is over at `v3.1-working` which is a lot ahead the master branch so I've implemented it there. Thanks!

PR #215, opened 2020-03-25
Use SPDX Identifier in README

oxr463:
See: https://spdx.org/licenses/Apache-2.0.html


FireFart:
Hi,
thanks but I think this is already enough documented

PR #251, opened 2020-11-07
Goroutine And Tip

blackcrw:
# 1. Line: 182 Goroutine
There is no need to call "Add()" twice and you can only call it once and inform that there will be two Goroutines.

Note: From what I learned from Goroutine, in this case there is no need to have two calls. I don't know if they did it for any specific reason.

# 2 Variables - Tip
I thought about changing the names of the variables, instead of just a letter, putting a real name, more readable and understandable.

### DIFF
--- a/cli/gobuster.go
+++ b/cli/gobuster.go
@@ -179,10 +179,8 @@ func Gobuster(prevCtx context.Context, opts *libgobuster.Options, plugin libgobu
 		MaxCharsWritten: 0,
 	}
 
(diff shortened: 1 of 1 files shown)


OJ:
Thanks for this, I'll merge in the next version 👍

FireFart:
This was done for readability and to be clear about it. There is no real benefit of `Add(2)` instead, I prefer the readability here. Thanks anyways for your contribution!

PR #259, opened 2020-12-21
Check if the user has given the FUZZ keyword or not (GoBuster fuzz)

shelld3v:
Check if the user has given the FUZZ keyword or not (GoBuster fuzz)


FireFart:
Hi,
thanks for the patch. We currently do all user input validation in the `cmd` package. Can you please move this check there?

shelld3v:
Okay!

Issue #271, opened 2020-12-31
Setting Host Header via -H Doesn't Set it in Requests


ryohare:
```bash
gobuster dir -u http://localhost:8000 -w list.txt -H 'Host: sdf'
```

```bash
sudo tcpdump -i lo -nn -A port 8000 | grep Host
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on lo, link-type EN10MB (Ethernet), capture size 262144 bytes
Host: localhost:8000
Host: localhost:8000
Host: localhost:8000
Host: localhost:8000
Host: localhost:8000
Host: localhost:8000
Host: localhost:8000
```

PR #272, opened 2020-12-31
Make -H 'Host: my-server' set Host Header on Requests

ryohare:
Found myself needing to dirbust with a custom host header to hit the right vhost. Couldnt find it in the tool, so I figured I'd add it.

It the libgobuster http client is initialized with a host header, req.Host will be set from the header value stored in the struct.

Opened and issue for this: https://github.com/OJ/gobuster/issues/271.

### DIFF
--- a/libgobuster/http.go
+++ b/libgobuster/http.go
@@ -22,6 +22,7 @@ type HTTPHeader struct {
 type HTTPClient struct {
 	client           *http.Client
 	context          context.Context
+	host             string
 	userAgent        string
 	defaultUserAgent string
 	username         string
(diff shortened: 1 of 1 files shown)


FireFart:
Hi @ryohare ,
thanks but this is wrongly implemented and will break VHOST mode. Currently there is the `Host` field on `libgobuster.RequestOptions` which should be passed from the single modes instead.

FireFart:
Ah sorry missed the exception, this should work. Will add a few modifications and comments to it, thanks!

ryohare:
I saw the Host parameter being passed in for the VHost mode, so yeah, I added it as the else in the if. Parsed out the host in the constructor for minor perf increases.

One thing I considered but would defer to the team here was adding the host parameter as an argument, like user-agent is. Just went with it as a header to make it a less significant user facing change....

Issue #276, opened 2021-02-05
Unable to exclude status code


duraki:
Hii! For some reason, I can't exclude (or better say include) the status code that makes up issue with wildcards.

See below for example:

```
# => includes only HTTP 200 OK
$ gobuster dir -u https://www.example.com/ -w ~/util/SecLists/Discovery/Web-Content/directory-list-lowercase-2.3-medium.txt -s 200
===============================================================
Gobuster v3.1.0
by OJ Reeves (@TheColonial) & Christian Mehlmauer (@firefart)
===============================================================
[+] Url:                     https://www.example.com/
[+] Method:                  GET
[+] Threads:                 10
[+] Wordlist:                ~/util/SecLists/Discovery/Web-Content/directory-list-lowercase-2.3-medium.txt
[+] Negative Status codes:   404
[+] User Agent:              gobuster/3.1.0
[+] Timeout:                 10s
[+] // Missing "Status Code"
===============================================================
2021/02/05 15:35:27 Starting gobuster in directory enumeration mode
===============================================================
Error: the server returns a status code that matches the provided options for non existing urls. https://www.example.com/da8306e5-10c0-416a-9c63-4f125bc96878 => 302 (Length: 26). To continue please exclude the status code, the length or use the --wildcard switch
```

Why does it even check for 302 if I only included 200 OK? Why is there no information about status code(s) I included in the stdout tty banner?

sloftis2016:
I've submitted a pull request (Update dir.go #286) to fix this issue. You can make the changes yourself to 2 lines of code to fix this behavior.

FireFart:
This works as expected. Everything except 404 is handled a "good hit". The Pre check tries some random urls that do not exist and every one of those result in  a 302 so there is no real way to detect this.
So you need to update the -b switch with 404,302 or exclude the length to continue

ZackInMA:
Will this ever be fixed? Still an issue in 3.1.0

Error: the server returns a status code that matches the provided options for non existing urls. https://acf51fdf1ef3226bc0c107b5001d0039.web-security-academy.net/5a5ad98b-471c-4dee-a868-dbf7b56088d5 => 302 (Length: 0). To continue please exclude the status code, the length or use the --wildcard switch

Admonstrator:
> Will this ever be fixed? Still an issue in 3.1.0
> 
[...]

PR #286, opened 2021-03-02
Update dir.go

sloftis2016:
There was a default value for Status-codes-blacklist (-b) flag. This causes the status-codes to be overwritten. Also - The Status-codes (-s) option wasn't populated with a default. Added those back into the option.

### DIFF
--- a/cli/cmd/dir.go
+++ b/cli/cmd/dir.go
@@ -142,8 +142,8 @@ func init() {
 	if err := addCommonHTTPOptions(cmdDir); err != nil {
 		log.Fatalf("%v", err)
 	}
-	cmdDir.Flags().StringP("status-codes", "s", "", "Positive status codes (will be overwritten with status-codes-blacklist if set)")
-	cmdDir.Flags().StringP("status-codes-blacklist", "b", "404", "Negative status codes (will override status-codes if set)")
+	cmdDir.Flags().StringP("status-codes", "s", "200,204,301,302,307,401,403", "Positive status codes (will be overwritten with status-codes-blacklist if set)")
+	cmdDir.Flags().StringP("status-codes-blacklist", "b", "", "Negative status codes (will override status-codes if set)")
 	cmdDir.Flags().StringP("extensions", "x", "", "File extension(s) to search for")
 	cmdDir.Flags().BoolP("expanded", "e", false, "Expanded mode, print full URLs")
 	cmdDir.Flags().BoolP("no-status", "n", false, "Don't print status codes")


FireFart:
Please see https://github.com/OJ/gobuster/issues/276, this is working as intended

sloftis2016:
This is not working as intended. 
If you run the tool in this manor:
# gobuster dir -u http://10.10.10.1 -w /usr/share/wordlist/common.txt 
What status-codes are intended to be returned by default? Without setting -s or -b flags?

I believe the intent based on the documentation is for -s (whitelist) to include  (default "200,204,301,302,307,401,403").

The way the code is currently written. There is a default value for -b ("404"). Which means the -s option will never work unless you use the -b flag with an empty string "". Maybe that is the intent but it is a departure from past builds.

`# gobuster dir -u https://postman:10000 -x cgi,sh **-s "204,301,304"** -k -w /usr/share/wordlists/dirbuster/directory-list-2.3-medium.txt -t 20                                                                                             1 ⨯
Gobuster v3.1.0
by OJ Reeves (@TheColonial) & Christian Mehlmauer (@firefart)
[+] Url:                     https://postman:10000
[+] Threads:                 20
[+] Wordlist:                /usr/share/wordlists/dirbuster/directory-list-2.3-medium.txt
[+] **Negative Status codes:   404**
[+] User Agent:              gobuster/3.1.0
[+] Extensions:              sh,cgi
2021/03/02 08:47:13 Starting gobuster in directory enumeration mode
Error: the server returns a status code that matches the provided options for non existing urls. https://postman:10000/463eecef-41a2-4061-9194-8713f4ac245e => 200 (Length: 4561). To continue please exclude the status code or the length
                                                                                                                                                                                                                                              
┌──(root💀mia)-[/usr/local/go/bin]
└─# gobuster dir -u https://postman:10000 -x cgi,sh **-s "204,301,304"** **-b ""** -k -w /usr/share/wordlists/dirbuster/directory-list-2.3-medium.txt -t 20                                                                                       1 ⨯
Gobuster v3.1.0
by OJ Reeves (@TheColonial) & Christian Mehlmauer (@firefart)
[+] Url:            https://postman:10000
[+] Method:         GET
[+] Threads:        20
[+] Wordlist:       /usr/share/wordlists/dirbuster/directory-list-2.3-medium.txt
**[+] Status codes:   204,301,304**
[+] User Agent:     gobuster/3.1.0
[+] Extensions:     cgi,sh
[+] Timeout:        10s
2021/03/02 08:47:25 Starting gobuster in directory enumeration mode
`

FireFart:
I've added better error messages when both of the status code parameters are provided

duraki:
I agree w/ @sloftis2016. This makes absolute no sense in application flow. There is no reason to close this without proper resolving of #276- as in; the flow is different when -b flag is empty or whether is unset (uses default), which in that case overrides the `-s` argument. From the UX perspective, there is no way end-user would understand this flow.

FireFart:
There is currently a solution, you only can set one of both. If you think there is a better solution to this, please make a proposal here so we can discuss this.

ZackInMA:
@FireFart 
The solution does not work if you specifically want to see 302 status codes..

FireFart:
It works. You are setting 302 as a good status code but all non found urls return a 302 so this will result in a lot of false positives. So you either need to exclude 302 or accept the warning and filter out the fp manually

KunZizinho (commented):
update

lonley-art:
Error: error on running gobuster: unable to connect to https://10.10.10.1:10000/: Get "https://10.10.10.1:10000/": context deadline exceeded (Client.Timeout exceeded while awaiting headers)




what can i do?

AndresG991:
How do I access to that file so I can change it?

PR #294, opened 2021-05-15
Added basic colored output support for gobusterdir

setersora:
Hello.

This PR implements an option to colorize the output of `gobuster dir` as requested in #115. I've used [gookit/color](https://github.com/gookit/color) to apply coloring because it's lightweight, doesn't require any additional packages except itself and seems to work well across different operating systems.

One important thing is that I've found the progress bar's paddings break the bar if current character position is behind the terminal's width. It appears not only with my patch applied but in current master build as well, so I've decided to remove those paddings.

Also please pay attention that the option is supposed to be turned on with a command line flag as was requested in previous similar PR (https://github.com/OJ/gobuster/pull/122#pullrequestreview-217536152).


firefart:
coloring is now implemented in the `dev` branch

PR #304, opened 2021-08-05
Added status code filter to vhost mode (-s -b)

patrickhener:
I added the filtering of status code (negative and positive) to the vhost modus according to dir modus.


firefart:
thanks this is already implemented on the dev branch

PR #311, opened 2021-10-21
Google Cloud Storage (GCS) bucket scanning

nlykkei:
Added support for Google Cloud Storage (GCS) bucket scanning. The scanning lists all public (anonymous access) buckets


firefart:
Thanks! I've added it to the dev branch for now to test it with the other upcoming changes

PR #321, opened 2022-01-01
Dev

albertobregliano:
I simplified a couple of functiond and added a test


firefart:
this does not add any value at all, sorry

PR #322, opened 2022-01-17
Additional Tip

AbhijithAJ:
When I run this script, I frequently check if it is complete. 
So I have developed a [intimations](https://github.com/AbhijithAJ/intimations) module which will intimate me via a telegram message once it is done.

Hope this will help others.


firefart:
there are multiple tools to do that and it's not related to gobuster so we will kepp these things out of the readme

PR #336, opened 2022-05-11
Exclude Length Range for Directory Enumeration

randomactsofsecurity:
Provides an additional argument to be used with `--exclude-length` in order to exclude a range of content lengths. Occasionally, I've had content length enumeration come back with many slight differences due to cookies, response body caching values, UUIDs randomly generated on each request, that I want to bulk exclude, such as:
 
```
1905
2200
2199
2197
```

However, I don't want to specify every single one of these in the `--exclude-length` argument.

Instead, I added `--exclude-length-range` which can be used to provide a base value which will setup a Min and Max off of the base `--exclude-length` value.

For example, if you had an incoming content length of 2000 that you want to exclude all content lengths from 1400 - 2600, you would setup the gobuster arguments as the following:

`gobuster dir -u <url> --exclude-length 2000 --exclude-length-range 600`

This should also work alongside the `-b` arguments to combo block lengths and status codes

### DIFF
--- a/cli/cmd/dir.go
+++ b/cli/cmd/dir.go
@@ -127,7 +127,12 @@ func parseDirOptions() (*libgobuster.Options, *gobusterdir.OptionsDir, error) {
 
 	plugin.ExcludeLength, err = cmdDir.Flags().GetIntSlice("exclude-length")
 	if err != nil {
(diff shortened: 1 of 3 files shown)


firefart:
thanks this is already implemented in the latest versions

Issue #337, opened 2022-05-18
Headers option changing case


Freccia:
While using option `--headers` I noticed it changed the case of the header's name.

Example request :
![image](https://user-images.githubusercontent.com/15819880/169069597-18ebe10b-748f-4f47-9956-c0c60774fdad.png)

Result : 
![image](https://user-images.githubusercontent.com/15819880/169069884-2505a41e-ffad-4d34-9da1-8b9fd1960f77.png)

Now, the API I am using doesn't like it, since it's case sensitive, so I really need it to use the case I am giving.

Freccia:
I found the issue. Proposing pull-request #338 to fix it.

Apparently [http.Header](http://godoc.org/net/http#Header) [Add](http://godoc.org/net/http#Header.Add) and [Set](http://godoc.org/net/http#Header.Set) methods canonicalize the header name when adding values to the header map.

There is a trick to fix this, as stated [here](https://stackoverflow.com/questions/26351716/how-to-keep-key-case-sensitive-in-request-header-using-golang)

firefart:
thanks this is now implemented on the dev branch

PR #338, opened 2022-05-18
Fixing case sensitive Headers option

Freccia:
Fixing #337 

Reference :
https://stackoverflow.com/questions/26351716/how-to-keep-key-case-sensitive-in-request-header-using-golang


firefart:
I would prefer this as an extra option, either a case insensitive list or a boolean indicating that the headers should not be normalized to not alter the default behavior which currently works for most webservers. please also use the dev branch for pull requests as that‘s the current development version

firefart:
thanks this is now implemented on the dev branch

PR #403, opened 2023-03-17
Added "wordlist-offset" flag to resume from specific position in the wordlist

mmgordon82:
## Summary
Added a new feature, that allows users to resume from a specific position in the wordlist file by specifying the `--wordlist-offset <int>` flag. This is a very non-intrusive and easy-to-use state-resuming mechanism. **When using huge wordlists, this feature is simply a must.**

## Motivation
While using Gobuster, I encountered a problem where I had to interrupt it to add filters. When I restarted the tool, it had to start from the beginning of the wordlist file, resulting in tedious re-iterations of previously tested URLs. This made it inefficient and time-consuming.

In some more advanced brute-force programs, the user can continue from an ongoing state (which happens a lot - from forgetting a flag to sudden internet interruptions). For example, [epi052/feroxbuster](https://github.com/epi052/feroxbuster) creates a state file on the current directory and a willing user may set `--resume-from <state-file>` and save time by not re-iterating the same elements repeatedly. I find this mechanism rather cumbersome because it may create many state files (this feature is on by default)

## Changes Made
To address this problem, I added a new flag, `--wordlist-offset <n: int>`, to the tool's command-line interface. This flag allows users to skip the first `n` elements in the wordlist file, thereby resuming the brute-force from that point. I also added a check for where the `n` is larger than the number of elements.

#### Command:
```bash
gobuster dir -u http://example.com/ -w wordlist.txt --wordlist-offset 400
```
#### Output:
```
Gobuster v3.5
[+] Wordlist:                .\wordlist.txt
[+] Negative Status codes:   404
[+] User Agent:              gobuster/3.5
2023/03/18 00:34:26 Starting gobuster in directory enumeration mode
2023/03/18 00:34:26 Skipping the first 400 elements...
Progress: 401 / 156211 (0.25%)
```

## Additional Information
I have tested this feature thoroughly on my local machine, and it works as expected.

I believe this feature will greatly enhance the tool's functionality and usability, and I hope it will be considered for inclusion in the next release.

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -88,6 +88,20 @@ func parseGlobalOptions() (*libgobuster.Options, error) {
 		return nil, fmt.Errorf("wordlist file %q does not exist: %w", globalopts.Wordlist, err2)
 	}
 
+	offset, err := rootCmd.Flags().GetInt("wordlist-offset")
+	if err != nil {
+		return nil, fmt.Errorf("invalid value for wordlist-offset: %w", err)
+	}
+
+	if offset < 0 {
+		return nil, fmt.Errorf("wordlist-offset must be bigger or equal to 0")
+	}
+	globalopts.WordlistOffset = offset
+
+	if globalopts.Wordlist == "-" && globalopts.WordlistOffset > 0 {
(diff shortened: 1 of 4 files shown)


firefart:
awesome thanks!

PR #477, opened 2023-12-30
Update Go version

veqqq:
The jump to 1.21 makes things a tad faster, without breaking anything (all tests passed etc.)


firefart:
Thanks but that’s already done in the dev branch

PR #495, opened 2024-04-20
chore: use errors.New to replace fmt.Errorf with no parameters

ChengenH:


firefart:
is there any real world benefit to this? if not I prefer the current style

ChengenH:
> is there any real world benefit to this? if not I prefer the current style

Just one more step to format, So if you need to format your error message fmt.Errorf(), if you don't errors.New().
Just a recommendation.

PR #506, opened 2024-05-17
Fix "local error: tls: no renegotiation" for self-signed certs

fkasler:
Some servers with self-signed certs response cause a "local error: tls: no renegotiation" error. To fix, I added "Renegotiation: tls.RenegotiateOnceAsClient" to the TLS config per this solution:

https://stackoverflow.com/questions/57420833/tls-no-renegotiation-error-on-http-request

Seems to fix the problem.


firefart:
Thanks I made this optional using a cli option so it's still disabled by default. You can try it on the dev branch if you like

PR #521, opened 2024-07-31
Exclude status code in VHOST

diego-tella:
I created a code that will filter a certain status code passed to the vhost.
image

`./gobuster vhost -u http://saori.hc/ -t 50 -w /usr/share/wordlists/subdomains.txt --append-domain --exclude-length 300-310 --exclude-code 301`

If I pass `--exclude-code 301` as an example it will not return any response that has this status code.

image

Example with `--exclude-code 400`

### DIFF
--- a/cli/cmd/vhost.go
+++ b/cli/cmd/vhost.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"log"
+	"strconv"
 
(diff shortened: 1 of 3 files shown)


firefart:
Thanks but this is already implemented in the `dev` branch. You can try it using `go install github.com/OJ/gobuster/v3@dev`

PR #530, opened 2024-10-03
Add exclude-hostname-length flag to dynamically adjust exclude-length

0xyy66:
This PR introduces the *--exclude-hostname-length* flag for the vhost option to Gobuster, allowing users to dynamically adjust the *exclude-length* value based on the length of the hostname (fuzzing word) in the response. This feature is useful for cases where webservers return the subdomain or hostname in the response, causing the response length to vary.
 
 ## Key Changes:
- New vhost flag *--exclude-hostname-length*
- Simple change in exclude length control logic: `if (found && !v.options.ExcludeLengthParsed.Contains(int(size)-wordLength))` (subtracts the hostname length from the returned size, always matching the static size entered by the user)

## PoC
In the following screenshot, the response size is variable because of the subdomain included in the response. The problem could be bypassed by submitting a range but it would be less precise.

gobuster_no_hostname_exclude

By submitting the static size to *--exclude-length* (283 in this case) and the *--exclude-hostname-length* flag, Gobuster skips all the previous false positives.

gobuster_hostname_exclude

### DIFF
--- a/cli/cmd/vhost.go
+++ b/cli/cmd/vhost.go
@@ -80,6 +80,11 @@ func parseVhostOptions() (*libgobuster.Options, *gobustervhost.OptionsVhost, err
 		return nil, nil, fmt.Errorf("invalid value for domain: %w", err)
 	}
 
+	pluginOpts.ExcludeHostnameLength, err = cmdVhost.Flags().GetBool("exclude-hostname-length")
+	if err != nil {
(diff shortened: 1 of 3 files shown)


firefart:
@0xyy66 could you please rebase the pr? thanks!

0xyy66:
@firefart should be all set now. Let me know if I missed anything!

firefart:
Thanks! Will be included in the next release

Issue #533, opened 2024-10-16
Progress goes over 100%


ManuelBerrueta:
Using the latest release at this time (v3.6) for Linux. I was running the tool to enumerate virtual hosts for a lab and the tool is going way over the 100% mark:
![image](https://github.com/user-attachments/assets/9af17ce4-5250-4077-923a-08c6db652ff2)

firefart:
Can you please try the dev branch and see if the problem exists there too?

go install github.com/OJ/gobuster/v3@dev

ManuelBerrueta:
Hello, I tried to install it and got an issue with the library: `/usr/libexec/gcc/x86_64-linux-gnu/13/cc1: error while loading shared libraries: /lib/x86_64-linux-gnu/libisl.so.23: invalid ELF header`. I would usually explore more about the issue, but time constraints!

NITHINSPACETIME:
> Hello, I tried to install it and got an issue with the library: `/usr/libexec/gcc/x86_64-linux-gnu/13/cc1: error while loading shared libraries: /lib/x86_64-linux-gnu/libisl.so.23: invalid ELF header`. I would usually explore more about the issue, but time constraints!

this error occurs while the library file /lib/x86_64-linux-gnu/libisl.so.23 is corrupted or invalid i guess... To solve this error try checking if this file exists : `ls -l /lib/x86_64-linux-gnu/libisl.so.23`
 
if it does then , try insatlling : `sudo apt update
sudo apt install --reinstall libisl23` 

if anything happenes ask !!!

firefart:
@Freeedim can you please try the `dev` branch? `go install github.com/OJ/gobuster/v3@dev`

There are a few updates to the progress handling in this version, curious if this problem still exists in the new code

Freeedim:
Sorry, I thought I had answered this --'
I am in a situation where I would rather not install the dev version. Maybe later... Thanks for your attention anyway.

firefart:
this should be fixed in v3.7.0

Issue #480, opened 2024-01-13
Pattern flag adds to the number of requests


DozyDoh:
When using -p to provide a file with patterns to be tested, the regular command (without p) is still processed.  This adds to the number of requests which might be unnecessary.  It would be nice when using -p to only process queries per line in the pattern file, and skip the regular query

husnain-ce:
Yes, It should have to be like that.

Issue #405, opened 2023-03-31
Patterns do not count in total percent calculation


cweiske:
I used a pattern file with 3 lines, and gobuster's progress bar went up to 300%.

Maybe the total URL count calculation does not take the pattern file into account.

Issue #364, opened 2022-11-02
Unexpected behavior in dns + patterns


secu77:
Hello,

I was trying the dns mode combined with the "patterns", but I'm not getting the expected results:

- I'm trying to fuzz subdomains of westeurope.cloudapp.azure.com.
- I'm using a basic wordlist with words like: dev, prod, test, etc.
- I intend to use a pattern list containing patters like: `<org>-{GOBUSTER}`, producing a result like: `acme-dev.westeurope.cloudapp.azure.com`

Under this context, I form the command to launch which is as follows: `gobuster dns -d westeurope.cloudapp.azure.com -t 1 -w wordlist.txt -p patterns.txt -o results.txt -v`

wordlist.txt
```
dev
test
prod
testing
demo
demoapp
beta
alpha
```
patterns.txt
```
acme-{GOBUSTER}
acme-{GOBUSTER}dev
acme-{GOBUSTER}build
acme-{GOBUSTER}demo
{GOBUSTER}-acme
```
But, when fuzzing starts, I notice in the results subdomains like: "dev.westeurope.cloudapp.azure.com" or "test.westeurope.cloudapp.azure.com".

By enabling the verbose, I can see that the patterns are being used, but **how can I prevent DNS queries that do not use the pattern from being made?**

![image_2022-11-02_10-43-51](https://user-images.githubusercontent.com/23587075/199458479-75cc3416-e153-4e2c-a416-4b54003e1446.png)

Best regards and thanks in advance

firefart:
Hi,
the pattern is currently in addition to the wordlist. Will look into adding a switch to disable this behaviour.

ethanannane:
I think this can be closed now.

PR #548, opened 2025-01-29
Fix progress with patterns and backup discovery

### LINKED ISSUE #298: --discover-backup not working as advertised, looks for files even if they weren't found
I noticed that without the `--discover-backup` flag, I had 800k requests lined up in the progress bar, but with the `--discover-backup` option, I had ~6 million. Reading the help text, it says that backup files are only enumerated if a file is found. Given that, wouldn't it make sense to display only the 800k requests in the progress bar, and update the amount of requests when files get found?

I'm using Kali's version of gobuster.

```
└─$ gobuster version               
3.1.0
```

I initially thought the progress bar would maybe jump to 100% when all 800k requests have been completed without a single found file.

However, after playing around with it, it seems the help text is wrong. `--discover-backup` will search for files even if they haven't been found.

Proof: (look at the time stamps and how long it took to run the scans - 3 seconds vs 20 seconds)

```
┌──(x90slide㉿kali)-[~/…/jeeves/testing/autorecon/scans]
└─$ gobuster dir -u http://jeeves.htb:80/ -w /tmp/small.txt --discover-backup --no-tls-validation --status-codes "200,204,301,302,307,403,500" | tee "/tmp/tcp_80_http_gobuster_words_combined.txt"        
Gobuster v3.1.0
by OJ Reeves (@TheColonial) & Christian Mehlmauer (@firefart)
[+] Url:                     http://jeeves.htb:80/
[+] Threads:                 10
[+] Wordlist:                /tmp/small.txt
[+] Negative Status codes:   404
[+] User Agent:              gobuster/3.1.0
2021/06/17 18:25:38 Starting gobuster in directory enumeration mode
                              
2021/06/17 18:25:58 Finished
┌──(x90slide㉿kali)-[~/…/jeeves/testing/autorecon/scans]
└─$ gobuster dir -u http://jeeves.htb:80/ -w /tmp/small.txt  --no-tls-validation --status-codes "200,204,301,302,307,403,500" | tee "/tmp/tcp_80_http_gobuster_words_combined.txt"     
Gobuster v3.1.0
by OJ Reeves (@TheColonial) & Christian Mehlmauer (@firefart)
[+] Url:                     http://jeeves.htb:80/
[+] Threads:                 10
[+] Wordlist:                /tmp/small.txt
[+] Negative Status codes:   404
[+] User Agent:              gobuster/3.1.0
2021/06/17 18:26:12 Starting gobuster in directory enumeration mode
                            
2021/06/17 18:26:15 Finished

```

The help entry says:

```
 -d, --discover-backup                 Upon finding a file search for backup files
```

Given that not a single file was found, I would have expected the directory enumeration to take the same amount of time on both occasions.

acammack:
Hi all, this resolves the issues where patterns were not counted towards possible attempts, file extension searching was not applied to patterns in `GobusterDir`, and backups were being brute-forced for the entire wordlist. These all had pretty tangled fixes so I lumped them into this PR.

```
$  ./gobuster dir -p patterns.txt -w testwords.txt -x .php,.html -db -u https://testing.test
Gobuster v3.7
by OJ Reeves (@TheColonial) & Christian Mehlmauer (@firefart)
[+] Url:                     https://testing.test
[+] Threads:                 10
[+] Wordlist:                testwords.txt
[+] Patterns:                patterns.txt (2 entries)
[+] Negative Status codes:   404
[+] User Agent:              gobuster/3.7
[+] Extensions:              php,html
Starting gobuster in directory enumeration mode
/index.html           (Status: 200) [Size: 179]
/index.html.bak       (Status: 200) [Size: 179]
Progress: 33 / 33 (100.00%)
Finished
$ wc -l testwords.txt patterns.txt 
 3 testwords.txt
 3 patterns.txt
 5 total
```

Note the 33 total attempts for:

```
3 words * 3 patterns * (2 file extensions + no file extension) = 27 original attempts
plus 6 backup styles for the original find of "index.html"
```

~Since it's in this wheelhouse, I'm open to also changing the pattern generation and wordlist behavior to either have a flag for the exclusion of the bare words from the wordlist that's either on or off by default or even make patterns always supersede the wordlist and make users add a plain `{GOBUSTER}` pattern if they want the original word in addition to the patterns. I'm not sure which would be the most useful, though.~

Looking through the old issues, it seems that (1) filtering the whole wordlist through the patterns and (2) applying the discovery patterns exactly once would be the most consistent behavior.

Fixes #298, #405, #480, #533, #319, #364

There are still a couple of things to do before this is ready, but I wanted to get this open for some eyes:
- [x]  I need to track successes from success pattern generation to keep it from infinitely generating requests against malicious services/with misconfiguration, possibly with a configurable chain generation limit
- [x] I want to add a global flag for post-find pattern expansion, so that if the user desired everything could be checked with patterns after a successful find

### DIFF
--- a/libgobuster/libgobuster.go
+++ b/libgobuster/libgobuster.go
@@ -266,13 +266,19 @@ ListenForMore:
 			break ListenForMore
 		case successGuess := <-successChan:
 			// Add more guesses based on the results of previous attempts
-			// TODO: add the option for arbitrary patterns based on successful finds
 			if successGuess.discoverOnSuccess {
-				additionalWords := g.plugin.AdditionalSuccessWords(successGuess.word)
-				if len(additionalWords) > 0 {
-					g.Progress.IncrementTotalRequests(len(additionalWords))
+				discoverWords := g.plugin.AdditionalSuccessWords(successGuess.word)
+				if len(discoverWords) > 0 {
+					g.Progress.IncrementTotalRequests(len(discoverWords))
 					feederGroup.Add(1)
-					go g.feeder(feederCtx, guessChan, additionalWords, false, &feederGroup)
+					go g.feeder(feederCtx, guessChan, discoverWords, false, &feederGroup)
+				}
+
+				patternDiscoverWords := g.processDiscoverPatterns(successGuess.word)
+				if len(patternDiscoverWords) > 0 {
+					g.Progress.IncrementTotalRequests(len(patternDiscoverWords))
+					feederGroup.Add(1)
+					go g.feeder(feederCtx, guessChan, patternDiscoverWords, false, &feederGroup)
 				}
 			}
 		case <-time.After(200 * time.Millisecond):
@@ -312,11 +318,21 @@ func (g *Gobuster) processPatterns(word string) []string {
 		return nil
 	}
 
-	//nolint:prealloc
-	var pat []string
-	for _, x := range g.Opts.Patterns {
-		repl := strings.ReplaceAll(x, PATTERN, word)
-		pat = append(pat, repl)
+	return g.applyPatterns(word, g.Opts.Patterns)
+}
+
+func (g *Gobuster) processDiscoverPatterns(word string) []string {
+	if g.Opts.DiscoverPatternFile == "" {
+		return nil
+	}
+
+	return g.applyPatterns(word, g.Opts.DiscoverPatterns)
+}
+
+func (g *Gobuster) applyPatterns(word string, patterns []string) []string {
+	pat := make([]string, len(patterns))
+	for i, x := range patterns {
+		pat[i] = strings.ReplaceAll(x, PATTERN, word)
 	}
 	return pat
 }
--- a/libgobuster/libgobuster.go
+++ b/libgobuster/libgobuster.go
@@ -41,38 +41,86 @@ func NewGobuster(opts *Options, plugin GobusterPlugin, logger *Logger) (*Gobuste
 	return &g, nil
 }
 
-func (g *Gobuster) worker(ctx context.Context, wordChan <-chan string, wg *sync.WaitGroup) {
+func (g *Gobuster) worker(ctx context.Context, wordChan <-chan string, moreWordsChan chan<- string, wg *sync.WaitGroup) {
 	defer wg.Done()
 	for {
 		select {
 		case <-ctx.Done():
 			return
-		case word, ok := <-wordChan:
-			// worker finished
-			if !ok {
-				return
-			}
-			g.Progress.incrementRequests()
+		case word := <-wordChan:
 
 			wordCleaned := strings.TrimSpace(word)
 			// Skip empty lines
 			if len(wordCleaned) == 0 {
+				g.Progress.incrementRequests()
 				break
 			}
 
 			// Mode-specific processing
-			err := g.plugin.ProcessWord(ctx, wordCleaned, g.Progress)
+			res, err := g.plugin.ProcessWord(ctx, wordCleaned, g.Progress)
 			if err != nil {
 				// do not exit and continue
 				g.Progress.ErrorChan <- fmt.Errorf("error on word %s: %w", wordCleaned, err)
-				continue
 			}
+
+			if res != nil {
+				g.Progress.ResultChan <- res
+				select {
+				case <-ctx.Done():
+					return
+				case moreWordsChan <- wordCleaned:
+				}
+			}
+
+			g.Progress.incrementRequests()
 
 			select {
 			case <-ctx.Done():
 			case <-time.After(g.Opts.Delay):
 			}
 		}
+	}
+}
+
+func feed(ctx context.Context, wordChan chan<- string, words []string) {
+	for _, w := range words {
+		select {
+		// need to check here too otherwise wordChan will block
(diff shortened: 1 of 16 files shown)


firefart:
Nice thanks already! One thing I noticed is that we should close all channels when we are done sending on them

acammack:
> One thing I noticed is that we should close all channels when we are done sending on them

Hmm, is that necessary here?  By the time we would `close` the `wordChan` and `moreWordsChan` channels, all the other goroutines referring to them will have finished and `close` won't have anyone to notify. The resources for the channels still won't be freed up until after `Run()` returns either way. I guess it wouldn't hurt anything either, but I feel it would muddy the intent of the code by implying that there might be consumers that need to be notified that the channels are done.

acammack:
Yes, all the producers and consumers are terminated by contexts now and verified with wait groups and request counts. I think this also fixes a few minor issues that have already been closed, like #541

firefart:
Awesome, so just put the PR from draft to normal once you are ready so I can have a test. Thanks so far!

acammack:
This should be good to test. I think the changes in here will help with adding machine readable results and a pause/state file, though I'll leave those to the future. I think we could also make our string manipulation more efficient, but that's also another PR.

libgobuster/libgobuster.go
```diff
@@ -41,33 +52,40 @@ func NewGobuster(opts *Options, plugin GobusterPlugin, logger *Logger) (*Gobuste
 	return &g, nil
 }
 
-func (g *Gobuster) worker(ctx context.Context, wordChan <-chan string, wg *sync.WaitGroup) {
+func (g *Gobuster) worker(ctx context.Context, guessChan <-chan *Guess, successChan chan<- *Guess, wg *sync.WaitGroup) {
 	defer wg.Done()
 	for {
+		// Prioritize stopping when the context is done
 		select {
```
firefart, line 59: I think this select is not necessary as the context is handled in the below select?

libgobuster/libgobuster.go
```diff
@@ -76,10 +94,93 @@ func (g *Gobuster) worker(ctx context.Context, wordChan <-chan string, wg *sync.
...
+			// Skip empty lines removing expected work
+			g.Progress.IncrementTotalRequests(-1 * wordlist.guessesPerLine)
+			continue
+		}
+
+		if len(g.Opts.Patterns) > 0 {
+			for _, w := range g.processPatterns(word) {
+				guess := &Guess{word: w, discoverOnSuccess: true}
+				select {
+				case <-ctx.Done():
```
firefart, line 150: I think this select is also not necessary

libgobuster/libgobuster.go
```diff
@@ -143,56 +234,81 @@ func (g *Gobuster) Run(ctx context.Context) error {
...
+	feederGroup.Add(1)
+	go g.feedWordlist(feederCtx, guessChan, wordlist, &feederGroup)
 
-Scan:
-	for scanner.Scan() {
+ListenForMore:
+	for {
+		// Prioritize stopping when the context is done
 		select {
 		case <-ctx.Done():
```
firefart, line 266: I think this select is also not necessary

libgobuster/libgobuster.go
acammack: In the Go spec, `select` is specified as:

> If one or more of the communications can proceed, a single one that can proceed is chosen via a uniform pseudo-random selection.

So while `<-ctx.Done()` worked as an escape-hatch to blocking writes using just one `select` statement, there are internal states that would lead to more work being queued up and executed than anticipated, with the work potentially being unbounded. This is made more likely with our use of buffered channels. The only way to prioritize select clauses is with extra tiers of `select`, and we need duplicate `<-ctx.Done()`s in each `select` so that we don't end up blocking on a channel write. While lacking deterministic shutdown wasn't a huge problem with the existing code, it will be handy for implementing pause/restart.

firefart: yes but you have an empty select that returns on a context done or continues. In the following select you also check for a done context so the previous select is not necessary as there are no operations between the context checks

acammack: The catch is the second `select` won't always choose `ctx.Done()` if there are guesses buffered in `guessChan`, it's a coin flip if both channels are available for reading. Consider the following program:
```go
package main

import "fmt"

func main() {
        ones := make(chan int, 10)
        twos := make(chan int, 10)

        for range 10 {
                ones <- 1
                twos <- 2
        }

        for range 20 {
                select {
                case <-ones:
                        fmt.Println("Got a one!")
                case <-twos:
                        fmt.Println("Got a two!")
                }
        }
}
```
It's output is something like:
```
$ go run simultaneous_receives.go 
Got a two!
Got a two!
Got a two!
Got a two!
Got a one!
Got a one!
Got a two!
Got a one!
Got a two!
Got a one!
Got a two!
Got a two!
Got a two!
Got a two!
Got a one!
Got a one!
Got a one!
Got a one!
Got a one!
Got a one!
```

firefart: thanks for the clarification ! missed that it’s just a single select

firefart:
Thanks! Will test it on the dev branch with the other changes that are staged for the next version

PR #594, opened 2025-09-04
Dev

firefart:


ayounilimited:
gobuster -u http://fakebank.thm -w wordlist.txt dir

PR #619, opened 2025-12-02
Update method flag description in CLI

atalatable:
There is a copy pasting error in the help menu for the method flag.


firefart:
thanks!

PR #626, opened 2026-01-03
Add --json/-j flag to output results in JSON format

timothyleung1:
- JSON output requires --output flag and writes to file only
- Implement ResultToJSON() method for all result types:
  * gobusterdir: path, status_code, size, location, headers
  * gobusterdns: subdomain, IPs, CNAME
  * gobusterfuzz: word, path, status_code, size, location, headers
  * gobustervhost: vhost, status_code, size, location, headers
  * gobusters3: found, bucket_name, status, URL
  * gobustergcs: found, bucket_name, status, URL
  * gobustertftp: filename, size, error_message
- File output formats results as valid JSON array with indentation
- Results only written to file, not printed to console when using JSON

### DIFF
--- a/cli/gobuster.go
+++ b/cli/gobuster.go
@@ -25,33 +25,79 @@ func resultWorker(g *libgobuster.Gobuster, filename string, wg *sync.WaitGroup) 
 
 	var f *os.File
 	var err error
+	var firstResult = true
+
 	if filename != "" {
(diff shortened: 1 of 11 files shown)


timothyleung1:
Related to #75

timothyleung1:
Will have a look when i have time this week

PR #632, opened 2026-01-19
Feat/autocalibration

crypt0g30rgy:


firefart:
Please provide more context about this PR and what your intentions are

crypt0g30rgy:
Gobuster Enhancements: Multi-Target, Autocalibrate, and Improved Output

Details the recent changes implemented in Gobuster to support multiple targets via a list, automatic wildcard calibration, and improved output file handling based on list scanning.

1. Multi-Target Support (-l / --list)
Users can now provide a file containing multiple target URLs using the -l or --list flag.

Implementation: The dir mode now parses the list file and iterates over each URL sequentially.
Robustness: If a scan for one target fails, Gobuster logs the error and continues with the next target in the list.
Improved Output: When running with multiple targets, "Expanded" mode (-e) is automatically enabled. This ensures that the full URL is displayed for each result, making it clear which target a result belongs to.

2. Autocalibrate (--ac / --autocalibrate)
The new --ac flag automates the detection and exclusion of wildcard responses.

How it works: During the pre-scan phase, Gobuster sends a request to a random GUID-based path. If the server returns a response that matches the "success" criteria (which would normally result in a Wildcard Error), the content length of this response is automatically added to the exclude-length list.
Benefit: This removes the manual step of identifying wildcard lengths, streamlining the scanning process against servers that return custom 404 pages with 200 OK status codes.

3. Improved Output Handling (Auto-Append)

Refined how the -o (output) flag behaves during multi-target runs to ensure no data is lost.

One-time Truncation: When starting a run with -l, the output file is truncated only once at the beginning of the entire process.
Automatic Appending: Subsequent targets automatically append their results to the same output file.
Logic: This is handled by forcing the append flag to true internally for each target run after the initial setup.

Intentions and Impact
The goal i was hoping to achieve with these changes is to make Gobuster more efficient for bulk scanning and more resilient to tricky server configurations. By automating wildcard detection and unifying output management, we reduce the manual configuration overhead for security researchers and penetration testers.

also i had hoped to first full merge this changes on my fork but i only understood today that i was merging on both my fork and the master of this... hoping this can be implemented into the project in the future, thanks.

PR #634, opened 2026-02-17
Fix --exclude-hostname-length feature

byt3loss:
## Issue description
In v3.8.2, the `--exclude-hostname-length` feature stopped working due to an incorrect variable reference.

The hostname length was calculated using `len(subdomain)` instead of `len(word)`, resulting in wrong filtering behavior.

gobuster_not_working

## Root cause
A regression introduced in v3.8.2 caused `hostnameLength` to reference the `subdomain` variable instead of the original `word` value.

## Fix
`hostnameLength` is now correctly assigned using `len(word)`, restoring the expected filtering behavior.

### DIFF
--- a/gobustervhost/gobustervhost.go
+++ b/gobustervhost/gobustervhost.go
@@ -135,7 +135,7 @@ func (v *GobusterVhost) ProcessWord(ctx context.Context, word string, progress *
 		subdomain = word
 	}
 	if v.options.ExcludeHostnameLength {
-		hostnameLength = len(subdomain)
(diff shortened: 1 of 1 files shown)


firefart:
the subdomain variable contains the whole domain which is correct, I guess your server prints out only the hostname without the whole domain? I think we should support both, and not one or another

byt3loss:
Got it, that makes sense 👍

My original reasoning was based on the fuzzing word only, since it represents the dynamic portion reflected in the 302 responses.
But referencing the full hostname works as well.

